PR #1, opened 2025-01-21
Move executorch export to optimum-executorch

echarlaix:
Following https://github.com/huggingface/optimum/pull/2090#issuecomment-2566266913

ExecuTorch export (introduced in https://github.com/huggingface/optimum/pull/2090) is moved in `optimum-executorch` see https://github.com/huggingface/optimum/pull/2165


setup.py
```diff
@@ -0,0 +1,70 @@
...
+# Ensure we match the version set in optimum/executorch/version.py
+filepath = "optimum/executorch/version.py"
+try:
+    with open(filepath) as version_file:
+        (__version__,) = re.findall('__version__ = "(.*)"', version_file.read())
+except Exception as error:
+    assert False, "Error: Could not open '%s' due %s\n" % (filepath, error)
+
+INSTALL_REQUIRE = [
+    # "optimum~=1.24",
```
michaelbenayoun, line 15: Why is it commented?

michaelbenayoun (approved):
LGTM!
Thanks for handling this

setup.py
echarlaix: needs to install optimum from source, we can upgrade to optimum v1.24 once released

guangy10:
@echarlaix do you mind explaining the motivation of moving to optimum-executorch? I remember you briefly mentioned that during out meetup at HF Paris office but can't recall the details. I want to make sure such a move will not potentially decrease the visibility executorch to HF users. Thanks!

optimum/commands/export/executorch.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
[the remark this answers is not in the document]
guangy10: Yeah, that's ok

echarlaix:
@guangy10 we are planning to keep in the optimum repo only things which are common to all subpackages and to move everything else in specific repos (the same will happen for anything onnx related for example). I understand your point concerning visibility, something we can do to make sure this is not the case is to clearly redirect users to this repo in the optimum and transformers documentation, what do you think ?

guangy10:
> @guangy10 we are planning to keep in the optimum repo only things which are common to all subpackages and to move everything else in specific repos (the same will happen for anything onnx related for example). I understand your point concerning visibility, something we can do to make sure this is not the case is to clearly redirect users to this repo in the optimum and transformers documentation, what do you think ?

Thanks for the clarification. With such move, is it possible for us to get more fine-grained metrics/stats? For example, compare DAU/downloads/checkout across different optimum supported backends, e.g. onnx, executorch, tflite, etc. Track inference per model. Maybe other metrics, do you know?

Issue #5, opened 2025-01-28
Fix required python version


guangy10:
Minimal required python version is 3.10 per [ExecuTorch setup](https://github.com/pytorch/executorch/blob/main/pyproject.toml)

echarlaix:
Fixed in https://github.com/huggingface/optimum-executorch/pull/8

PR #8, opened 2025-01-28
Fix minimal python version

guangy10:
Fix for #5

ExecuTorch supported python 3.10+ and tested with 3.10, 3.11 and 3.12 explicitly.


guangy10:
cc: @echarlaix for review. BTW, we should probably bring the Issue/PR template to this repo.

setup.py
```diff
@@ -51,9 +51,9 @@
         "Intended Audience :: Science/Research",
         "Operating System :: OS Independent",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
+        "Programming Language :: Python :: 3.12",
```
guangy10, line 56: We can remove this line if 3.12 is not supported by optimum: https://github.com/huggingface/optimum/blob/main/setup.py#L123-L126

echarlaix (approved):
LGTM, thanks @guangy10 !

PR #13, opened 2025-02-03
Enable loading model from hub that has already been converted 

echarlaix:
* Enable loading model from the hub that have already been converted to executorch like [this model](https://huggingface.co/optimum-internal-testing/tiny-random-llama/tree/executorch)

* Infer whether the model should be exported or not by checking whether a `.pte` file is present when loading the model

```python
from optimum.executorch import ExecuTorchModelForCausalLM

model_id = "optimum-internal-testing/tiny-random-llama"
model = ExecuTorchModelForCausalLM.from_pretrained(model_id, revision="executorch")
```

### DIFF
--- a/optimum/executorch/modeling.py
+++ b/optimum/executorch/modeling.py
@@ -127,19 +127,13 @@ class ExecuTorchModelForCausalLM(OptimizedModel):
         file_name: Optional[str] = None,
         **kwargs,
     ) -> "ExecuTorchModelForCausalLM":
-        model_path = Path(model_id)
+        if isinstance(model_id, Path):
+            model_id = model_id.as_posix()
+
         default_file_name = file_name or "model.pte"
 
-        if local_files_only and not model_path.is_dir():
-            object_id = str(model_id).replace("/", "--")
-            cached_model_dir = os.path.join(cache_dir, f"models--{object_id}")
-            refs_file = os.path.join(os.path.join(cached_model_dir, "refs"), revision or "main")
-            with open(refs_file) as f:
-                _revision = f.read()
-            model_path = Path(os.path.join(cached_model_dir, "snapshots", _revision))
-
         pte_files = find_files_matching_pattern(
-            str(model_path),
+            model_id,
             _FILE_PATTERN,
             glob_pattern="**/*.pte",
             subfolder=subfolder,
@@ -148,7 +142,7 @@ class ExecuTorchModelForCausalLM(OptimizedModel):
         )
 
         if len(pte_files) == 0:
-            raise FileNotFoundError(f"Could not find any ExecuTorch model file in {model_path}")
+            raise FileNotFoundError(f"Could not find any ExecuTorch model file in {model_id}")
         if len(pte_files) == 1 and file_name and file_name != pte_files[0].name:
             raise FileNotFoundError(f"Trying to load {file_name} but only found {pte_files[0].name}")
 
@@ -168,12 +162,12 @@ class ExecuTorchModelForCausalLM(OptimizedModel):
                 f"Loading the file {file_name} in the subfolder {subfolder}."
             )
 
-        if model_path.is_dir():
-            model_path = subfolder
+        if os.path.isdir(model_id):
+            model_id = subfolder
             subfolder = ""
 
         model_cache_path = cls._cached_file(
-            model_path=model_path,
+            model_path=model_id,
             token=token,
             revision=revision,
             force_download=force_download,
--- a/optimum/executorch/modeling.py
+++ b/optimum/executorch/modeling.py
@@ -124,38 +124,57 @@ class ExecuTorchModelForCausalLM(OptimizedModel):
         force_download: bool = False,
         local_files_only: bool = False,
         cache_dir: str = HUGGINGFACE_HUB_CACHE,
+        file_name: Optional[str] = None,
         **kwargs,
     ) -> "ExecuTorchModelForCausalLM":
-        """
-        Load a pre-trained ExecuTorch model from a local directory or hosted on the HF hub.
 
-        Args:
-            model_id (`Union[str, Path]`):
-                Path to the directory containing the ExecuTorch model file (`model.pte`).
-            config (`PretrainedConfig`, *optional*):
-                Configuration of the pre-trained model.
-            token (`Optional[Union[bool,str]]`, defaults to `None`):
-                The token to use as HTTP bearer authorization for remote files. If `True`, will use the token generated
-                when running `huggingface-cli login` (stored in `huggingface_hub.constants.HF_TOKEN_PATH`).
-            subfolder (`str`, defaults to `""`):
-                In case the relevant files are located inside a subfolder of the model repo either locally or on huggingface.co, you can
-                specify the folder name here.
-            revision (`str`, defaults to `"main"`):
-                Revision is the specific model version to use. It can be a branch name, a tag name, or a commit id.
-            cache_dir (`Optional[str]`, defaults to `None`):
-                Path indicating where to store cache. The default Hugging Face cache path will be used by default.
-            force_download (`bool`, defaults to `False`):
-                Whether or not to force the (re-)download of the model weights and configuration files, overriding the
-                cached versions if they exist.
-            local_files_only (`Optional[bool]`, defaults to `False`):
-                Whether or not to only look at local files (i.e., do not try to download the model).
-
-        Returns:
-            `ExecuTorchModelForCausalLM`: The initialized ExecuTorch model.
-
-        """
         model_path = Path(model_id)
-        default_file_name = "model.pte"
+        default_file_name = file_name or "model.pte"
+
+        if local_files_only:
+            object_id = str(model_id).replace(os.sep, "--")
+            cached_model_dir = os.path.join(cache_dir, f"models--{object_id}")
+            refs_file = os.path.join(os.path.join(cached_model_dir, "refs"), revision or "main")
+            with open(refs_file) as f:
+                _revision = f.read()
+            model_dir = os.path.join(cached_model_dir, "snapshots", _revision)
+        else:
+            model_dir = str(model_id)
+
+        pte_files = find_files_matching_pattern(
+            model_dir,
+            _FILE_PATTERN,
+            glob_pattern="**/*.pte",
+            subfolder=subfolder,
+            token=token,
+            revision=revision,
+        )
+
+        model_path = Path(model_dir)
+        if len(pte_files) == 0:
+            raise FileNotFoundError(f"Could not find any ExecuTorch model file in {model_dir}")
+        if len(pte_files) == 1 and file_name and file_name != pte_files[0].name:
+            raise FileNotFoundError(f"Trying to load {file_name} but only found {pte_files[0].name}")
+
+        file_name = pte_files[0].name
+        subfolder = pte_files[0].parent
+
+        if len(pte_files) > 1:
+            for file in pte_files:
+                if file.name == default_file_name:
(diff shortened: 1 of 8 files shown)


guangy10:
@echarlaix thank you for adding support to load converted and cached ExecuTorch model from hub. We are thinking of supporting same use cases. I'd like to get your insights how we can collaborate on it together. To make the discussion easier I opened a new GitHub Issue here: https://github.com/huggingface/optimum-executorch/issues/15

optimum/executorch/modeling.py
```diff
@@ -456,3 +389,83 @@ def text_generation(
...
+                revision=revision,
+            )
+            _export = len(pte_files) == 0
+            if _export ^ export:
+                if export:
+                    logger.warning(
+                        f"The model {model_id} was already converted to the ExecuTorch IR but got `export=True`, the model will be converted to ExecuTorch once again. "
+                        # "Don't forget to save the resulting model with `.save_pretrained()`"
+                    )
+                    _export = True
```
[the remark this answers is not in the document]
guangy10: Thanks for preserving it for usability and adding the warning message

optimum/executorch/modeling.py
```diff
@@ -113,107 +111,32 @@ def forward(
...
     @classmethod
-    def from_pretrained(
+    def _from_pretrained(
         cls,
-        model_name_or_path: Union[str, Path],
-        export: bool = True,
-        recipe: str = "",
-        config: "PretrainedConfig" = None,
-        subfolder: str = "",
+        model_id: Union[str, Path],
```
guangy10, line 119: nit: It can still be a path to a local directory, right? Isn't the name confusing?

optimum/executorch/modeling.py
```diff
@@ -113,107 +111,32 @@ def forward(
...
+        config: PretrainedConfig,
+        token: Optional[Union[bool, str]] = None,
         revision: Optional[str] = None,
-        cache_dir: str = HUGGINGFACE_HUB_CACHE,
+        subfolder: str = "",
         force_download: bool = False,
         local_files_only: bool = False,
-        use_auth_token: Optional[Union[bool, str]] = None,
-        token: Optional[Union[bool, str]] = None,
+        cache_dir: str = HUGGINGFACE_HUB_CACHE,
```
guangy10, line 126: Just to check, is this pointing to the local cache, i.e. `~/.cache/huggingface/hub/` on my devserver?

optimum/executorch/modeling.py
```diff
@@ -113,107 +111,32 @@ def forward(
...
         **kwargs,
     ) -> "ExecuTorchModelForCausalLM":
         """
-        Load a pre-trained ExecuTorch model.
+        Load a pre-trained ExecuTorch model from a local directory or hosted on the HF hub.
```
guangy10, line 130: This is great! Will continue reviewing this PR, but by reading this comment I assume this PR is enabling load cached .pte from both a local directory and remotely hosted on HF hub

optimum/executorch/modeling.py
```diff
@@ -226,54 +149,79 @@ def _from_pretrained(
...
         """
-        full_path = os.path.join(f"{model_dir_path}", "model.pte")
-        model = _load_for_executorch(full_path)
-        logging.info(f"Loaded model from {full_path}")
-        logging.debug(f"{model.method_meta('forward')}")
-        return cls(
-            model=model,
-            config=config,
+        model_path = Path(model_id)
+        default_file_name = "model.pte"
```
guangy10, line 158: Yeah, the default model name looks good to me

optimum/executorch/modeling.py
```diff
@@ -226,54 +149,79 @@ def _from_pretrained(
...
+        model_cache_path = cls._cached_file(
+            model_path=model_path,
+            token=token,
+            revision=revision,
+            force_download=force_download,
+            cache_dir=cache_dir,
+            file_name=default_file_name,
+            subfolder=subfolder,
+            local_files_only=local_files_only,
         )
```
guangy10, line 169: 👍

optimum/executorch/modeling.py
```diff
@@ -226,54 +149,79 @@ def _from_pretrained(
...
+        model = _load_for_executorch(model_cache_path)
+        logging.info(f"Loaded model from {model_cache_path}")
 
-    def _save_pretrained(self, save_directory):
```
guangy10, line 248: I think the `_save_pretrained` is implemented already, when we export to ExecuTorch and save the pte to a local fs. I think we can just move the logic to this API, to provide HF users a consistent experience when working with ExecuTorch backend

optimum/executorch/modeling.py
```diff
@@ -226,54 +149,79 @@ def _from_pretrained(
...
+    def _cached_file(
+        model_path: Union[Path, str],
+        token: Optional[Union[bool, str]] = None,
+        revision: Optional[str] = None,
+        force_download: bool = False,
+        cache_dir: Optional[str] = None,
+        file_name: Optional[str] = None,
+        subfolder: str = "",
+        local_files_only: bool = False,
+    ):
```
guangy10, line 185: Do you mind adding the API documentation for it?

optimum/executorch/modeling.py
```diff
@@ -226,54 +149,79 @@ def _from_pretrained(
...
+        model_path = Path(model_path)
+        # locates a file in a local folder and repo, downloads and cache it if necessary.
+        if model_path.is_dir():
+            model_cache_path = os.path.join(model_path, subfolder, file_name)
+        else:
+            model_cache_path = hf_hub_download(
```
guangy10, line 191: IIUC it will try loading from a local directory only if the model_id is a local path, i.e. `from_pretrained(model_id=my/local/path/to/executorch/model/)`. If the `model_id` is a hub id `meta-llama/Llama-3.2-1B`, it will download the cached one from hub. There is a case when the pte is already cached locally when calling `from_pretrained(model_id=meta-llama/Llama-3.2-1B)`, shouldn't it prioritize to return the local cached one instead?

optimum/executorch/modeling.py
```diff
@@ -336,18 +274,14 @@ def _export(
...
-            token=token,
-            local_files_only=local_files_only,
-            force_download=force_download,
-        )
+    def _save_pretrained(self, save_directory):
+        """
+        Saves a model weights into a directory, so that it can be re-loaded using the
+        [`from_pretrained`] class method.
+        """
+        raise NotImplementedError
```
guangy10, line 284: See my comment above https://github.com/huggingface/optimum-executorch/pull/13/files#r1947337461.

But it can be done in a separate PR.

tests/models/test_modeling.py
```diff
@@ -16,52 +16,68 @@
...
+from optimum.utils.file_utils import find_files_matching_pattern
 
 
 class ExecuTorchModelIntegrationTest(unittest.TestCase):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-    @slow
-    @pytest.mark.run_slow
     def test_load_model_from_hub(self):
```
guangy10, line 35:
```suggestion
    def test_load_cached_model_from_hub(self):
```

tests/models/test_modeling.py
```diff
@@ -16,52 +16,68 @@
...
         self.assertIsInstance(model.model, ExecuTorchModule)
 
-    @slow
-    @pytest.mark.run_slow
-    def test_load_model_from_local_path(self):
-        from optimum.exporters.executorch import main_export
+    def test_load_et_model_from_hub(self):
+        model_id = "optimum-internal-testing/tiny-random-llama"
+
+        model = ExecuTorchModelForCausalLM.from_pretrained(model_id, revision="executorch", recipe="xnnpack")
```
guangy10, line 45: I'm confused what are the difference w/ and w/o `revision="executorch"`?

tests/models/test_modeling.py
```diff
@@ -16,52 +16,68 @@
...
+        self.assertIsInstance(model, ExecuTorchModelForCausalLM)
+        self.assertIsInstance(model.model, ExecuTorchModule)
 
-        model_id = "NousResearch/Llama-3.2-1B"
-        task = "text-generation"
+    def test_load_model_from_local_path(self):
```
guangy10, line 49:
```suggestion
    def test_load_cached_model_from_local_path(self):
```

guangy10 (commented):
Look through the PR but it's still unclear to me where exactly the cached model is fetched from hub. Let's taking the "meta-llama/Llama-3.2-1B" for example. 

Is it fetching the the cached model from https://huggingface.co/executorch-community/Llama-3.2-1B-Instruct assuming this is an entry like that?

Or it's fetching the cached model from https://huggingface.co/meta-llama/Llama-3.2-1B under "Files and versions" tab?

guangy10:
The logic that publishes pte to hub doesn't exist yet, which will be enabled in a separate PR, right?

optimum/executorch/modeling.py
echarlaix: yes exactly!

optimum/executorch/modeling.py
echarlaix: yes by default, `cache_dir` can be set to something else if needed

optimum/executorch/modeling.py
echarlaix: we need to save the resulting pte file in `_save_pretrained` https://github.com/huggingface/optimum-executorch/blob/38c978229d3dff3598fa81d6d25b16c8f9cbfab0/optimum/executorch/modeling.py#L253

can be done in a following PR

optimum/executorch/modeling.py
echarlaix: sure I can add it in a following PR

optimum/executorch/modeling.py
echarlaix: Yes! I used `model_id` to match what we currently have for `OptimizedModel` 

https://github.com/huggingface/optimum/blob/856b25267ca686f05706cd44f0ebb669618cf302/optimum/modeling_base.py#L336

and `ORTModel`

https://github.com/huggingface/optimum/blob/856b25267ca686f05706cd44f0ebb669618cf302/optimum/onnxruntime/modeling_ort.py#L142

but if you think this is not clear we can think about an alternative

tests/models/test_modeling.py
echarlaix: if no pte files detected then `export` will be set to `True` and the model will be converted to ExecuTorch on-the-fly, after this the user can either save the resulting model locally with `.save_pretrained(save_dir)` or directly push it on the hub with `.push_to_hub(repo_id)` (both method needs to be implemented, let me know if you're interested by tackling this in a following PR)

echarlaix:
> 
> 
> Look through the PR but it's still unclear to me where exactly the cached model is fetched from hub. Let's taking the "meta-llama/Llama-3.2-1B" for example.
> 
> Is it fetching the the cached model from https://huggingface.co/executorch-community/Llama-3.2-1B-Instruct assuming this is an entry like that?
> 
> Or it's fetching the cached model from https://huggingface.co/meta-llama/Llama-3.2-1B under "Files and versions" tab?

If the `model_id` is set to `meta-llama/Llama-3.2-1B`  then  https://huggingface.co/meta-llama/Llama-3.2-1B 

> The logic that publishes pte to hub doesn't exist yet, which will be enabled in a separate PR, right?

can be integrated via a `push_to_hub` method in https://github.com/huggingface/optimum-executorch/pull/13#discussion_r1951238301

echarlaix:
Merging and we can add more features (saving exported model locally, pushing on the hub, docs) in following PRs cc @guangy10

PR #21, opened 2025-02-18
Add documentation workflows

echarlaix:


.github/workflows/upload_pr_documentation.yml
```diff
@@ -0,0 +1,16 @@
...
+    types:
+      - completed
+
+jobs:
+  build:
+    uses: huggingface/doc-builder/.github/workflows/upload_pr_documentation.yml@main
+    with:
+      package_name: optimum-executorch
+    secrets:
+      hf_token: ${{ secrets.HF_DOC_BUILD_PUSH }}
```
echarlaix, line 15: Needing the `HF_DOC_BUILD_PUSH` secret for this repo cc @glegendre01

.github/workflows/upload_pr_documentation.yml
```diff
@@ -0,0 +1,16 @@
+name: Upload PR Documentation
```
echarlaix, line 1: Not sure why but looks like this workflow is not triggered once "Build PR Documentation" completed

echarlaix:
@regisss let me know if we're good to merge so at least we have the documentation for https://github.com/huggingface/optimum/pull/2193 (I can remove the upload documentation workflow in the meantime and fix it later)

.github/workflows/upload_pr_documentation.yml
echarlaix: should be fixed with https://github.com/huggingface/optimum-executorch/pull/22 thanks @regisss :goat:

Issue #14, opened 2025-02-03
report regression in executorch-0.5.0


guangy10:
ExecuTorch 0.5.0 was released: https://github.com/pytorch/executorch/releases/tag/v0.5.0. However, this repo is new and there is no cross-repo integration tests, so there may be some regressions in 0.5.0. However, since 0.5.0 is out, there is no way to fix it back in 0.5.0, we can only target to get all known integration issues fixed by the upcoming `0.6.0` in Mar, 2025. The good thing is that we start pinning optimum-executorch to ExecuTorch CI in [Run Hugging Face models via Optimum on CI pytorch/executorch#8630](https://github.com/pytorch/executorch/pull/8630)

If you run into issue using executorch-0.5.0, you can be unblocked by trying either:
  - downgrade to 0.4.0 by running `pip install executorch==0.4.0`
  - install executorch from `main` branch.

guangy10:
Notice an issue when running Qwen2.5 with `executorch 0.5.0`. Works fine with 0.4.0

```
[tensor_util.h:713] Tensors do not match: numel=(2048,  1), dim=(1, 2)
[tensor_util.h:720]     size(0): (2048, 1)
[op_le.cpp:29] Check failed (tensors_have_same_shape(a, b)):
[method.cpp:1072] KernelCall failed at instruction 0:18 in operator aten::le.Tensor_out: 0x12
[method.cpp:1078] arg 0 with type id 1
[method.cpp:1078] arg 1 with type id 1
[method.cpp:1078] arg 2 with type id 1
[method.cpp:1078] arg 3 with type id 1
Traceback (most recent call last):
  File "/Users/guangyang/optimum-executorch/optimum/executorch/modeling.py", line 360, in text_generation
    generated_tokens = self.generate(
                       ^^^^^^^^^^^^^^
  File "/Users/guangyang/optimum-executorch/optimum/executorch/modeling.py", line 301, in generate
    logits = self.forward(
             ^^^^^^^^^^^^^
  File "/Users/guangyang/optimum-executorch/optimum/executorch/modeling.py", line 114, in forward
    return self.model.forward((input_ids, cache_position))[0]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: method->execute() failed with error 0x12
```

PR #27, opened 2025-02-25
Support checkpoint dtype via kwargs

guangy10:


guangy10:
The failure on Qwen model is a regression in `0.5.0` unfortunately (https://github.com/huggingface/optimum-executorch/issues/14), since there was no integration tests with ExecuTorch. This should have been fixed in `main` and will be available in 0.6.0 release next month. For now, I think we can just pin the executorch to a fixed version `0.4.0`?

setup.py
```diff
@@ -13,7 +13,7 @@
 
 INSTALL_REQUIRE = [
     "optimum~=1.24",
-    "executorch>=0.4.0",
+    "executorch==0.4.0",
```
guangy10, line 16: fyi https://github.com/huggingface/optimum-executorch/pull/27#issuecomment-2680157942

optimum/exporters/executorch/convert.py
```diff
@@ -87,6 +87,6 @@ def export_to_executorch(
     full_path = os.path.join(f"{output_dir}", "model.pte")
     with open(full_path, "wb") as f:
         executorch_prog.write_to_file(f)
-        logging.info(f"Saved exported program to {full_path}")
+        print(f"Saved exported program to {full_path}")
```
michaelbenayoun, line 90: Why?

optimum/commands/export/executorch.py
```diff
@@ -51,6 +51,13 @@ def parse_args_executorch(parser):
         default="xnnpack",
         help='Pre-defined recipes for export to ExecuTorch. Defaults to "xnnpack".',
     )
+    required_group.add_argument(
+        "--dtype",
+        type=str,
+        default="float32",
+        choices=["float32", "bfloat16", "float16"],
+        help='Data type of the pretrained checkpoint to load. Defaults to "float32".',
+    )
```
michaelbenayoun, line 60: It is not possible to infer that?

Issue #2, opened 2025-01-24
ExecuTorch documentation is broken on Hub


guangy10:
Documentation for ExecuTorch was under https://huggingface.co/docs/optimum/exporters/overview but can't find it any more. cc: @echarlaix

guangy10:
It looks like all executorch related docs are deleted in https://github.com/huggingface/optimum/pull/2165

echarlaix:
Hi @guangy10! We're planning to open a PR to add the documentation in `optimum-executorch` directly like for [optimum-neuron](https://github.com/huggingface/optimum-neuron/tree/main/docs) and others. This will introduce an executorch section in the main optimum documentation, would that option works for you ?

guangy10:
> Hi [@guangy10](https://github.com/guangy10)! We're planning to open a PR to add the documentation in `optimum-executorch` directly like for [optimum-neuron](https://github.com/huggingface/optimum-neuron/tree/main/docs) and others. This will introduce an executorch section in the main optimum documentation, would that option works for you ?

Is the documentation accessible from https://huggingface.co? Is it still under optimum?

guangy10:
I think there will be an entry for ExecuTorch and its subpages here: https://huggingface.co/docs/optimum/index

echarlaix:
> I think there will be an entry for ExecuTorch and its subpages here: https://huggingface.co/docs/optimum/index

Yes that's the plan!

guangy10:
@echarlaix Can we prioritize to bring the documentation back asap? We're planning to reach out to hardware partners, university students, etc. to encourage the contribution to this repo by expanding to support new models, recipes, etc. We need to point to them the documentation so that they can get started. Thanks! 

FYI @cbilgin

guangy10:
@echarlaix I don't see ExecuTorch is added back https://huggingface.co/docs/optimum/index. Do I miss anything?

echarlaix:
An executorch section will be added to the optimum documentation after https://github.com/huggingface/optimum/pull/2193 is merged

Issue #29, opened 2025-02-28
🐛 Bugbash #1: Help Improve optimum-executorch! 🚀


guangy10:
We're kicking off a **bugbash** for [`optimum-executorch`](https://github.com/huggingface/optimum-executorch)! This is your chance to test the repo, report issues, and provide feedback to help make it more robust and developer-friendly.  

## 🔥 How to Participate  
1. **Try it out** – Follow the setup instructions and test key workflows.  
2. **Check known issues** – Before reporting, review the [list of known issues](https://github.com/huggingface/optimum-executorch/issues?q=is%3Aissue+state%3Aopen+label%3Abug) to avoid duplication.  
3. **Report new issues** – If you find a bug, create a comment here or [open a new issue](https://github.com/huggingface/optimum-executorch/issues/new) with the `bugbash` label.  
4. **Share feedback** – Let us know if something is unclear, needs better documentation, or could be improved.  

## 🎯 Focus Areas  
We’re especially looking for feedback on:  
- Installation & setup easiness  
- General UX for Model export & inference using ExecuTorch  
- Documentation clarity  

#### Bugbash Path 1.1 Installation
- **How-to:** Follow [Step 1&2 in Quick Installation](https://github.com/huggingface/optimum-executorch/blob/37511c024280fe9734cefa33e04aaf73d446efe2/README.md#-quick-installation)
- **Expected outcomes:**
  - No installation errors
  - pip list to verify the required packages are installed: ➡ [required packages](https://github.com/huggingface/optimum-executorch/blob/main/setup.py#L14)

#### Bugbash Path 1.2 Export and Load in One Python API
- **How-to:** Follow [this section in README.md](https://github.com/huggingface/optimum-executorch/blob/37511c024280fe9734cefa33e04aaf73d446efe2/README.md#option-1-export-and-load-in-one-python-api)
- **Expected outcomes:**
  - Confirm successful export and inference for all supported models

#### Bugbash Path 1.3 Export and Load from Cache
- **How-to:** Follow [this section in README.md](https://github.com/huggingface/optimum-executorch/blob/37511c024280fe9734cefa33e04aaf73d446efe2/README.md#option-1-export-and-load-in-one-python-api)
- **Expected outcomes:**
  - First time will download the cached `pte` from Hugging Face Hub *(❌ it should not export from source)*
  - Rerun the API with the same model_id will directly load the cached `pte` from local fs *(❌ it should not redownload it from Hub)*

[...]

PR #31, opened 2025-03-05
Improve setup guide

guangy10:
- Moved more packages to be required to simplify the setup
- Added a section to setup virtual env
- Added a section to install key dependencies from source
- Explained API behavior when an ExecuTorch model is cached on Hub
- Added a list of supported models and recipes due to #2

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,18 +12,19 @@ except Exception as error:
     assert False, "Error: Could not open '%s' due %s\n" % (filepath, error)
 
 INSTALL_REQUIRE = [
-    "optimum~=1.24",
+    "accelerate>=0.26.0",
+    "datasets",
     "executorch>=0.4.0",
+    "optimum~=1.24",
+    "safetensors",
+    "sentencepiece",
+    "tiktoken",
     "transformers>=4.46",
 ]
 
 TESTS_REQUIRE = [
-    "accelerate>=0.26.0",
-    "pytest",
     "parameterized",
-    "sentencepiece",
-    "datasets",
-    "safetensors",
+    "pytest",
 ]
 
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -12,18 +12,19 @@
     assert False, "Error: Could not open '%s' due %s\n" % (filepath, error)
 
 INSTALL_REQUIRE = [
-    "optimum~=1.24",
+    "accelerate>=0.26.0",
+    "datasets",
```
echarlaix, line 16: why do we need `accelerate` and `datasets` as required dependencies ?

echarlaix (commented):
LGTM!

setup.py
guangy10: @echarlaix Yeah, required when running certain models, as reported by users here https://github.com/huggingface/optimum-executorch/issues/29#issuecomment-2691496917

I want to simplify the UX so that users can have most common deps installed by default.

echarlaix: I'd be in favor of keeping the number of required dependencies as low as possible (only keeping what's necessary). Could you extend on why datasets needs to be added? Also for accelerate I don't think this is mandatory (we could for example check if accelerate is available and depending on it how to set `low_cpu_mem_usage` when loading the model) https://github.com/huggingface/transformers/blob/v4.49.0/src/transformers/modeling_utils.py#L3612  wdyt ?

echarlaix: I don't want to block this PR so reverted in https://github.com/huggingface/optimum-executorch/pull/31/commits/7768d35a6ea8e80a081cf712d0efa647580944bc, would you mind opening a new PR for this so that we merge this PR asap and continue the discussion there?

echarlaix: also cc @michaelbenayoun who will likely review the second PR

PR #39, opened 2025-03-25
Version bump and fixes

guangy10:
The fix in https://github.com/pytorch/executorch/pull/9507 is merged, which will unblock us working on some new models in `transformers`. Bump and test the latest version and squeeze some improvements:
 - fix hard-coded `decoder_start_token_id`
 - to be compatible with the upcoming `executorch==0.6.0`
 - version bump so that we can work on many new `transformers` models from latest releases
 - minor bug fix for passing config
 - support phi4 e2e with customized config (no "longrope" support atm)  (causing OOM, so have to disable in CI atm)

### DIFF
--- a/optimum/executorch/modeling.py
+++ b/optimum/executorch/modeling.py
@@ -100,6 +100,8 @@ class ExecuTorchModelForSeq2SeqLM(OptimizedModel):
             self.vocab_size = self.et_decoder.run_method("get_vocab_size")[0]
         if "max_hidden_seq_length" in metadata:
             self.max_hidden_seq_length = self.et_decoder.run_method("max_hidden_seq_length")[0]
+        if "decoder_start_token_id" in metadata:
+            self.decoder_start_token_id = self.et_decoder.run_method("decoder_start_token_id")[0]
 
     def forward(
         self,
@@ -382,8 +384,9 @@ class ExecuTorchModelForSeq2SeqLM(OptimizedModel):
             )
             max_seq_len = self.max_cache_size
 
-        # Initialize with start token (0 for T5)
-        decoder_input_ids = torch.tensor([[0]], dtype=torch.long)
+        if not hasattr(self, "decoder_start_token_id"):
+            raise AttributeError("'decoder_start_token_id' is missing in the metadata of the PTE.")
+        decoder_input_ids = torch.tensor([[self.decoder_start_token_id]], dtype=torch.long)
         encoder_input_ids = input_ids
         encoder_outputs = None
         generated_ids = [0]
@@ -512,7 +515,7 @@ class ExecuTorchModelForCausalLM(OptimizedModel):
     def _from_pretrained(
         cls,
         model_id: Union[str, Path],
-        config: PretrainedConfig,
+        config: Optional[PretrainedConfig] = None,
         token: Optional[Union[bool, str]] = None,
         revision: Optional[str] = None,
         subfolder: str = "",
@@ -610,7 +613,7 @@ class ExecuTorchModelForCausalLM(OptimizedModel):
         cls,
         model_id: str,
         recipe: str,
-        config: PretrainedConfig,
+        config: Optional[PretrainedConfig] = None,
         token: Optional[Union[bool, str]] = None,
         revision: Optional[str] = None,
         cache_dir: str = HUGGINGFACE_HUB_CACHE,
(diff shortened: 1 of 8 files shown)


guangy10:
OOM when running Phi-4-mini (3.2B), no surprise as Llama3.2-3B causes OOM already. Disable this test on CI, but users can still convert this model using `ExecuTorchModelForCausalLM.from_pretrained()`.

guangy10:
cc @tugsbayasgalan for review as well

optimum/executorch/modeling.py
```diff
@@ -382,8 +384,7 @@ def generate(
             )
             max_seq_len = self.max_cache_size
 
-        # Initialize with start token (0 for T5)
-        decoder_input_ids = torch.tensor([[0]], dtype=torch.long)
+        decoder_input_ids = torch.tensor([[self.decoder_start_token_id]], dtype=torch.long)
```
guangy10, line 387: @michaelbenayoun FYI, addressed your comment in https://github.com/huggingface/optimum-executorch/pull/36

tests/models/test_modeling_phi4.py
```diff
@@ -0,0 +1,54 @@
...
+
+import pytest
+from executorch.extension.pybindings.portable_lib import ExecuTorchModule
+from transformers import AutoConfig, AutoTokenizer
+from transformers.testing_utils import slow
+
+from optimum.executorch import ExecuTorchModelForCausalLM
+
+
+@pytest.mark.skip(reason="Test Phi-4-mini (3.8B) will require runner to be configured with larger RAM")
```
guangy10, line 27: @michaelbenayoun can we add runners to the CI? Also add linux platform back?

optimum/executorch/modeling.py
michaelbenayoun: It seems that `self.decoder_start_token_id` exists only if `"decoder_start_token_id" in metadata`.
Should we define a default value for this attribute?
Otherwise we would need to check that `self` actually has this attribute in the forward or it could fail, right?

optimum/exporters/executorch/__main__.py
```diff
@@ -63,6 +65,8 @@ def main_export(
             The recipe to use to do the export, e.g. "xnnpack".
         output_dir (`Union[str, Path]`):
             Path indicating the directory where to store the generated ExecuTorch model.
+        config (`Optional[PretrainedConfig]`):
```
michaelbenayoun, line 68:
```suggestion
        config (`Optional[PretrainedConfig]`, defaults to `None`):
```

tests/models/test_modeling_phi4.py
michaelbenayoun: You mean runners that have more RAM? Let's sync offline on Slack about it.

About Linux platform, I'm all for it, it's just a matter of editing the github workflow right?

optimum/executorch/modeling.py
guangy10: Yeah, the metadata must be recorded to the PTE file to instruct the runtime for inference.So for encoder-decoder model which the `decoder_start_token_id` field is mandatory, if it's missing, we should raise an error.

tests/models/test_modeling_phi4.py
guangy10: @michaelbenayoun IIUC, the linux runners are causing OOM though the config seems to be fine. I can put a separate PR just adding the Linux to the workflow so we would know whether this issue is still legit, and minimize the chance of blocking this PR. How does it sound? And which linux runner should I use?

guangy10:
Addressed comments

tests/models/test_modeling_phi4.py
guangy10: @michaelbenayoun FYI, the Linux CI is kicked off here: https://github.com/huggingface/optimum-executorch/pull/40

tests/models/test_modeling_phi4.py
```diff
@@ -0,0 +1,54 @@
...
+    @pytest.mark.run_slow
+    def test_phi4_text_generation(self):
+        model_id = "microsoft/Phi-4-mini-instruct"
+        config = AutoConfig.from_pretrained(model_id)
+        # NOTE: To make the model exportable we need to set the rope scaling to default to avoid hitting
+        # the data-dependent control flow in _longrope_frequency_update. Alternatively, we can rewrite
+        # that function to avoid the data-dependent control flow.
+        if hasattr(config, "rope_scaling") and config.rope_scaling is not None:
+            config.rope_scaling["type"] = "default"
+        model = ExecuTorchModelForCausalLM.from_pretrained(model_id, recipe="xnnpack", config=config)
```
guangy10, line 42: cc @tugsbayasgalan. An example of load with customized config.

PR #42, opened 2025-03-28
Fix install from zsh and CI with executorch nightly

guangy10:
- Fixed `pip install .[tests]` for `zsh`
- Fixed CI to ensure the compatibility with executorch nightly (pinned to a specific version for stability as there is no guarantee all nightly build are good).
- Improved causal lm eval by utilizing the eager model to score perplexity value.


guangy10:
> LGTM, is it expected that the tests fail? It might be due to the change of the ExecuTorch version

Oh it looks like the failure is due to the generated texts are slightly different given different versions of export+executorch, but the outputs are still reasonable. I will need to refactor the test to loose the check a bit

guangy10:
@michaelbenayoun Looks like the 2 job failures are OOM, which happens sometimes on the 2B model. Can you help merge the PR?

tests/__init__.py
```diff
@@ -0,0 +1,14 @@
+# coding=utf-8
+# Copyright 2024 The HuggingFace Team. All rights reserved.
```
michaelbenayoun, line 2:
```suggestion
# Copyright 2025 The HuggingFace Team. All rights reserved.
```

michaelbenayoun (commented):
LGTM.
A few CIs fail, let's merge once they are passing

guangy10:
> LGTM. A few CIs fail, let's merge once they are passing

@michaelbenayoun There is nothing I can do with the OOM failure, exact same code/test, may pass in some runs but fails in others. We will need this PR to fix nightly integration, feel free to disable those tests if there is no other way to resolve the OOM issue

PR #44, opened 2025-04-04
Add latency measurement for text models

guangy10:
1. Add latency measurement for text models
  - Bring the ExecuTorch c++ runner measurements ([`stats.h`](https://github.com/pytorch/executorch/blob/main/extension/llm/runner/stats.h)) to `optimum-executorch` in python. Eventually we could remove the py impl and utilize it from extensions/llm via pybind.
  - Measure the latency of `ExecuTorchModelForSeq2SeqLM` and `ExecuTorchModelForCausalLM`

2. `modeling.py` refactor to reduce duplicate code:
 - Create a `ExecuTorchModelBase` that represent the ExecuTorch inference model, with implementation of export, load from cache/hub, etc. common methods. Defined abstract method `forward`, `generate`, etc. that each derived concrete class must implement
 - Making all supported ExecuTorch inference classes (`ExecuTorchModelForSeq2SeqLM`, `ExecuTorchModelForCausalLM`, etc.) lightweight and easy to scale

3. Added `efficientnet` model. Conditionally run only when executorch >= 0.6 (so it will run on the pinned nightly). Previously broken due to a bug a in XNNPACK, which has been fixed and picked to the upcoming `executorch 0.6` release.

### DIFF
--- a/.github/workflows/test_models.yml
+++ b/.github/workflows/test_models.yml
@@ -24,6 +24,7 @@ jobs:
           - deit
           - distilbert
           - dit
+          - efficientnet
           - focalnet
           - gemma
           - gemma2
@@ -35,8 +36,10 @@ jobs:
           - pvt
           - qwen2
           - roberta
+          - smollm
           - swin
           - t5
+          - vit
         executorch-version: ['0.4.0', 'nightly']
         python-version: ['3.10', '3.11', '3.12']
         os: [macos-15]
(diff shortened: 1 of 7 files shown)


michaelbenayoun (approved):
LGTM from what I can understand.
This is great work but could you please try to make separate PRs by topics, like here we could have: 
- A pr for the refactoring
- A pr for efficientnet
- A pr for stats computation

It is no big deal at all, but it makes things easier to review.

PR #48, opened 2025-04-08
Integration test with 0.6.0rc in CI

guangy10:
As titled, to ensure compatibility with upcoming `executorch==0.6.0` releaese.


guangy10:
Verified in CI that executorch==0.6.0 got installed and all tests finished as expected.

PR #55, opened 2025-04-14
Fix torch deps in CI

guangy10:


guangy10:
FYI, all CIs are green so I will go ahead merge this improvement.

PR #56, opened 2025-04-16
Support ExecuTorch 0.6.0

guangy10:
**THIS PR IS A PLACEHOLDER AND SHOULD NOT BE MERGED UNTIL 4/24/2025, AFTER THE `executorch 0.6.0` IS GENERAL AVAILABLE.**

This PR:
- bump up the min supported version of `executorch`
- install official package instead of RC

cc: @mergennachin @metascroy


guangy10:
Cancelled all GitHub Actions since the pkg is yet available. Will need to re-run on 4/24 before merge.

PR #60, opened 2025-04-29
Disclaimer for python-based perf report

guangy10:
To explain why we need the python-based measurement and how it should be interpreted/used.

image


guangy10:
cc: @jackzhxng for review

jackzhxng (approved):

PR #64, opened 2025-05-01
Run Gemma3 in CI

guangy10:
I think https://github.com/huggingface/optimum/issues/2127 is still a blocker. @michaelbenayoun @echarlaix, can you introduce using  `@require_read_token` in optimum?


guangy10:
Switch to use non-gated repos. Verified results in CI

PR #62, opened 2025-05-01
Introduce 8da4w quant for decoder-only text models

guangy10:
Initial efforts to introduce quantization to native Hugging Face models that are already supported in optimum-executorch. Start with decoder-only text models using "`8da4w`" for linear layers and int8 for embedding. 

Experiment the quantization configs with the following models:
 - `Qwen3-0.6B`
 -  `gemma-3-1b`
 - `HuggingFaceTB/SmolLM2-135M`

#### Example usage 
via `optimum-cli`
`optimum-cli export executorch --model Qwen/Qwen3-0.6B --task text-generation --recipe xnnpack --use_custom_sdpa --qlinear --qembedding --output_dir qwen3_8da4w_8we`

or use the `ExecuTorchModelForCausalLM.from_pretrained`.

`et_model = ExecuTorchModelForCausalLM.from_pretrained("./qwen3_8da4w_8we")`

Screenshot 2025-05-05 at 1 30 43 PM

#### `.pte` size comparison:
```
qwen3_8da4w_8we:
total 1035336
-rw-r--r--  506M  May  1 13:09 model.pte

qwen3_8da4w:
total 1944584
-rw-r--r--  950M  May  1 13:16 model.pte

qwen3_float16:
total 2937408
-rw-r--r--  1.4G  May  1 13:22 model.pte

qwen3_float32:
total 5873128
-rw-r--r--  2.8G  May  1 13:26 model.pte
```

PR #67, opened 2025-05-07
Support lowering quantized checkpoint from Hub

guangy10:
Users can lower a HF model with quantization by:
1. Load the fp32 checkpoint and apply the quantization, export and lowering recipe to generate the PTE, which is supported in https://github.com/huggingface/optimum-executorch/pull/62
2. Load a pre-quantized checkpoint and directly lower it to ExecuTorch via the exact same `from_pretrained` API. Behind the scene, it will bypass the quantization as the checkpoint has been quantized already. This path shows the composability of different recipes.

This PR is handling path 2 above.


guangy10:
```
E           ValueError: Failed to find class AOPerModuleConfig in any of the allowed modules: torchao.quantization, torchao.sparsity.sparse_api, torchao.prototype.quantization
```
Will need to upgrade the pinned torchao.

guangy10:
Rebased after https://github.com/huggingface/optimum-executorch/pull/68 is merged

optimum/executorch/passes/remove_padding_idx_embedding_pass.py
```diff
@@ -0,0 +1,22 @@
+import torch
+from executorch.exir.dialects._ops import ops as exir_ops
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
+class RemovePaddingIdxEmbeddingPass(ExportPass):
```
metascroy, line 6: Should we add tests for the pass?

metascroy (approved):

optimum/executorch/passes/remove_padding_idx_embedding_pass.py
guangy10: Yeah, so I will just copy over the test in your notebook

PR #69, opened 2025-05-22
Leverage update_cache op to reduce overhead from cache update

kimishpatel:
Summary:
This likely might be a short lived optimization where in future we can replace update_cache op with index_put_ op.
This is what original StaticCache does, however this requires cache transpose for custom_sdpa (which can also be fixed).

We will leverage custom cache for now, however in near future this should not be needed. This option however will allow us to bypass any transposes if the need continues

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/optimum/commands/export/executorch.py
+++ b/optimum/commands/export/executorch.py
@@ -28,7 +28,11 @@ if TYPE_CHECKING:
 def parse_args_executorch(parser):
     required_group = parser.add_argument_group("Required arguments")
     required_group.add_argument(
-        "-m", "--model", type=str, required=True, help="Model ID on huggingface.co or path on disk to load model from."
+        "-m",
+        "--model",
+        type=str,
+        required=True,
+        help="Model ID on huggingface.co or path on disk to load model from.",
     )
     required_group.add_argument(
         "-o",
@@ -58,6 +62,12 @@ def parse_args_executorch(parser):
         help="For decoder-only models to use custom sdpa with static kv cache to boost performance. Defaults to False.",
     )
     required_group.add_argument(
+        "--use_custom_kv_cache",
+        required=False,
+        action="store_true",
+        help="For decoder-only models to use custom kv cache for static cache that updates cache using custom op. Defaults to False.",
+    )
+    required_group.add_argument(
         "--qlinear",
         required=False,
         action="store_true",
@@ -84,6 +94,8 @@ class ExecuTorchExportCommand(BaseOptimumCLICommand):
         kwargs = {}
         if self.args.use_custom_sdpa:
             kwargs["use_custom_sdpa"] = self.args.use_custom_sdpa
+        if self.args.use_custom_kv_cache:
+            kwargs["use_custom_kv_cache"] = self.args.use_custom_kv_cache
         if self.args.qlinear:
             kwargs["qlinear"] = self.args.qlinear
         if self.args.qembedding:
--- a/optimum/executorch/attentions/custom_kv_cache.py
+++ b/optimum/executorch/attentions/custom_kv_cache.py
@@ -1,0 +1,256 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Any, Dict, Optional, Tuple, Union
+
+import torch
+
+
+try:
+    from transformers.cache_utils import StaticCache
+except ImportError:
+    # If transformers is not installed, raise an ImportError
+    try:
+        from transformers.cache_utils import StaticCache
+    except ImportError:
+        raise ImportError("transformers is not installed. Please install it to use StaticCache.")
+
+
+class ETCustomStaticCache(StaticCache):
+    """
+    Custom KV Cache implementation for ExecutorTorch that inherits from Hugging Face's StaticCache
+    but uses custom operations for cache updates similar to ExecutorTorch's CustomStaticCache.
+    """
+
+    def __init__(
+        self,
+        config,
+        max_batch_size: int,
+        max_cache_len: Optional[int] = None,
+        device: Union[torch.device, str, None] = None,
+        dtype: torch.dtype = torch.float32,
+        layer_device_map: Optional[Dict[int, Union[str, torch.device, int]]] = None,
+    ):
+        super().__init__(
+            config=config,
+            max_batch_size=max_batch_size,
+            max_cache_len=max_cache_len,
+            device=device,
+            dtype=dtype,
+            layer_device_map=layer_device_map,
+        )
+
+        # make sure layer_device_map is none
+        assert layer_device_map is None
+
+        # Clear existing caches
+        self.key_cache = []
+        self.value_cache = []
+
+        # Initialize cache buffers with our custom shape
+        cache_shape = (
+            self.max_batch_size,
+            self.max_cache_len,
+            self.num_key_value_heads,
+            self.head_dim,
+        )
+        assert device is None or device == "cpu", "Device must be None or 'cpu'"
+
+        for _ in range(config.num_hidden_layers):
+            self.new_layer_key_cache = torch.zeros(cache_shape, dtype=dtype, device="cpu")
+            self.new_layer_value_cache = torch.zeros(cache_shape, dtype=dtype, device="cpu")
+
+            self.key_cache.append(self.new_layer_key_cache)
+            self.value_cache.append(self.new_layer_value_cache)
+
+    def update(
+        self,
+        key_states: torch.Tensor,
+        value_states: torch.Tensor,
+        layer_idx: int,
+        cache_kwargs: Optional[Dict[str, Any]] = None,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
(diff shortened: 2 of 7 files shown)


optimum/executorch/attentions/custom_kv_cache.py
```diff
@@ -0,0 +1,257 @@
...
+            raise ImportError(
+                "ExecutorTorch custom operations are not available. "
+                "Please install executorch with custom operations support."
+            )
+
+    # Recursively replace KV caches
+    return _replace_with_et_custom_kv_cache(module, config, generation_config, cache_dtype)
+
+
+def _replace_with_et_custom_kv_cache(module, config, generation_config, cache_dtype):
```
guangy10, line 219: Can we move these to https://github.com/huggingface/optimum-executorch/tree/main/optimum/executorch/passes?

optimum/exporters/executorch/integrations.py
```diff
@@ -55,9 +56,34 @@ def export(self, input_ids=None, cache_position=None) -> Dict[str, ExportedProgr
...
 
             with torch.no_grad():
                 exported_program = exportable_module.export(example_input_ids, example_cache_position)
+                # Apply RemoveTransposes pass to remove
+                # any back-to-back transpose ops that are not needed
+                # e.g. output of update_cache is transposed and
+                # input to custom_sdpa is transposed.
+                from executorch.extension.llm.export.export_passes import (
+                    RemoveRedundantTransposes,
+                )
```
guangy10, line 79: This is causing problems in gemma3 CI. It can be easily resolved by adding 'tiktoken' to `TESTS_REQUIRE` in setup.py.
```
File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/optimum/exporters/executorch/__main__.py", line 138, in main_export
    return export_to_executorch(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/optimum/exporters/executorch/convert.py", line 77, in export_to_executorch
    executorch_progs = recipe_func(model, **kwargs)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/optimum/exporters/executorch/recipes/xnnpack.py", line 95, in export_to_executorch_with_xnnpack
    exported_progs = model.export()
                     ^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/optimum/exporters/executorch/integrations.py", line 77, in export
    from executorch.extension.llm.export.export_passes import (
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/executorch/extension/llm/export/__init__.py", line 7, in <module>
    from .builder import LLMEdgeManager
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/executorch/extension/llm/export/builder.py", line [38](https://github.com/huggingface/optimum-executorch/actions/runs/15214849869/job/42805259883?pr=69#step:5:39), in <module>
    from executorch.extension.llm.tokenizer.utils import get_tokenizer
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/executorch/extension/llm/tokenizer/utils.py", line 9, in <module>
    from executorch.examples.models.llama.tokenizer.tiktoken import Tokenizer as Tiktoken
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/executorch/examples/models/llama/tokenizer/tiktoken.py", line 23, in <module>
    import tiktoken
ModuleNotFoundError: No module named 'tiktoken'
```

However, note that the import deps are very ugly. It doesn't make sense that an extension module has dependency to examples/models/llama. There is no guarantee that examples will continue working and living in ET.

tests/models/test_modeling_qwen3.py
```diff
@@ -174,3 +175,38 @@ def test_qwen3_text_generation_with_custom_sdpa_8da4w_8we(self):
         gc.collect()
 
         self.assertTrue(check_causal_lm_output_quality(model_id, generated_tokens))
+
+    @slow
+    @pytest.mark.run_slow
+    @pytest.mark.skipif(
+        parse(transformers.__version__) < parse("0.52.0"),
+        reason="Only available on transformers >= 0.52.0",
+    )
```
guangy10, line 184: Why is it?

guangy10 (changes requested):
The perf boost seems significant! Thanks for adding the custom KV cache to HF models

optimum/commands/export/executorch.py
```diff
@@ -57,6 +61,12 @@ def parse_args_executorch(parser):
         action="store_true",
         help="For decoder-only models to use custom sdpa with static kv cache to boost performance. Defaults to False.",
     )
+    required_group.add_argument(
+        "--use_custom_kv_cache",
+        required=False,
+        action="store_true",
+        help="For decoder-only models to use custom kv cache for static cache that updates cache using custom op. Defaults to False.",
```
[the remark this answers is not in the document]
kimishpatel: No the two are independent. You dont need one to add the other.

optimum/executorch/attentions/custom_kv_cache.py
kimishpatel: oh didnt realize that existed. But I think this is a module swap not a graph rewrite pass. We should keep it here

tests/models/test_modeling_qwen3.py
kimishpatel: because of this https://fburl.com/wlu2s0u9

optimum/exporters/executorch/integrations.py
kimishpatel: ok I will add tiktoken. But I think it needs to be added to INSTALL_REQUIRE. Because otherwise you cannot use this scripts locally.

guangy10: > But I think it needs to be added to INSTALL_REQUIRE. Because otherwise you cannot use this scripts locally.

It's needed only when running it with specific models. I used to have a PR that trying to simplify the packages by just merging `TESTS_REQUIRE` with `INSTALL_REQUIRE`, but I think @echarlaix prefer to keep it separated. To be consistent I will just add it to `TESTS_REQUIRE` together with other packages like `sentencepiece`

guangy10: > also it is a bit of unfortunate code organization in et

Indeed. We should make it cleaner in ET codebase

tests/models/test_modeling_qwen3.py
guangy10: I see. https://github.com/huggingface/optimum-executorch/pull/70 is WIP, we will get rid of this check after it's merged. It's introduced due to FC (some features/models require pre-released version.)

@kimishpatel  BTW, you should fix the typo. It should be `4.52.0`.

optimum/executorch/attentions/custom_kv_cache.py
guangy10: Okay

optimum/commands/export/executorch.py
```diff
@@ -28,7 +28,11 @@
 def parse_args_executorch(parser):
     required_group = parser.add_argument_group("Required arguments")
     required_group.add_argument(
-        "-m", "--model", type=str, required=True, help="Model ID on huggingface.co or path on disk to load model from."
+        "-m",
+        "--model",
+        type=str,
+        required=True,
+        help="Model ID on huggingface.co or path on disk to load model from.",
```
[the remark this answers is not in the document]
guangy10: Yeah, guideline by hf/optimum. https://github.com/huggingface/optimum/blob/main/CONTRIBUTING.md#how-to-create-a-pull-request

tests/models/test_modeling_qwen3.py
kimishpatel: oh yeah my bad

optimum/exporters/executorch/integrations.py
kimishpatel: ok updated the test yml

tests/models/test_modeling_qwen3.py
```diff
@@ -174,3 +175,38 @@ def test_qwen3_text_generation_with_custom_sdpa_8da4w_8we(self):
         gc.collect()
 
         self.assertTrue(check_causal_lm_output_quality(model_id, generated_tokens))
+
+    @slow
+    @pytest.mark.run_slow
+    @pytest.mark.skipif(
+        parse(transformers.__version__) < parse("4.52.0"),
+        reason="Only available on transformers >= 4.52.0",
+    )
```
guangy10: BTW, I noticed this test is skipped in the CI https://github.com/huggingface/optimum-executorch/actions/runs/15429343459/job/43424006619?pr=69#step:5:876.

It's skipped because the installed version `4.52.0.dev0` (because it's installed from source in the CI as when I enabled gemma3, 4.52 is not released yet). To fix this you will need just compare the base version or compare version with dev0, example here https://github.com/huggingface/optimum-executorch/blob/34cece4f5819177c3196da10a8ccc47f053ee367/tests/models/test_modeling_gemma3.py#L43-L46

kimishpatel: ok let me fix that. BTW there are other test failures that seem unrelated

PR #70, opened 2025-05-28
Transformers version bump

guangy10:
To workaround the [bug](https://github.com/pytorch/pytorch/issues/150994) in `torch==2.7.0` (`executorch==0.6.0`), need to set `strict=False` for `export` when bumping Transformers to `4.52+`.

Pass custom inputs to export and enable dynamic shapes (for parallel prefill) will require export recipe from `transformers>=4.52.0`.

Fix the following error in phi-4 w/ quantized checkpoint
```
E           ValueError: Failed to find class ModuleFqnToConfig in any of the allowed modules: torchao.sparsity.sparse_api, torchao.prototype.quantization, torchao.quantization
```
will require a newer nightly of `torchao`.

### DIFF
--- a/.github/workflows/test_models.yml
+++ b/.github/workflows/test_models.yml
@@ -51,24 +51,18 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies for ExecuTorch
         run: |
+          pip install '.[tests]'
           if [ "${{ matrix.executorch-version }}" == "nightly" ]; then
-            export NIGHTLY_VERSION=dev20250507
+            export NIGHTLY_VERSION=dev20250523
             pip install executorch==0.7.0.${NIGHTLY_VERSION} \
               torch==2.8.0.${NIGHTLY_VERSION} \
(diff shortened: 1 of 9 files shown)


.github/workflows/test_models.yml
```diff
@@ -51,25 +51,19 @@ jobs:
...
+          pip install '.[tests]'
           if [ "${{ matrix.executorch-version }}" == "nightly" ]; then
-            export NIGHTLY_VERSION=dev20250507
+            export NIGHTLY_VERSION=dev20250523
             pip install executorch==0.7.0.${NIGHTLY_VERSION} \
               torch==2.8.0.${NIGHTLY_VERSION} \
               torchvision==0.22.0.${NIGHTLY_VERSION} \
               torchaudio==2.6.0.${NIGHTLY_VERSION} \
-              torchao==0.12.0.${NIGHTLY_VERSION} \
+              torchao==0.12.0.dev20250528 \
```
guangy10, line 61: There is no single nightly version that can align all these packages together so have to specify it separately.

PR #74, opened 2025-06-06
Simplify installation for dev

guangy10:
- Simplified the setup for dev, which typically require installing torch related packages from latest nightly and installing transformers from source.
- Updated instructions in the README.md.
- Use this script in CI.


README.md
```diff
@@ -135,7 +117,9 @@ print(generated_text)
 ## Supported Optimizations
 
 ### Custom Operators
-Supported using [custom SDPA](https://github.com/pytorch/executorch/blob/a4322c71c3a97e79e0454a8223db214b010f1193/extension/llm/README.md?plain=1#L40) with Hugging Face Transformers, boosting performance by 3x compared to default SDPA, based on tests with `HuggingFaceTB/SmolLM2-135M`.
+Supported using [**custom SDPA**](https://github.com/pytorch/executorch/blob/a4322c71c3a97e79e0454a8223db214b010f1193/extension/llm/README.md?plain=1#L40) with Hugging Face Transformers, boosting performance by **3x** compared to default SDPA, based on tests with `HuggingFaceTB/SmolLM2-135M`.
+
+Supported using **custom KV cache** that performs in-place cache update, boosting performance by **2.5x** compared to default static KV cache, based on tests with `HuggingFaceTB/SmolLM2-135M`.
```
guangy10, line 122: @kimishpatel Let me know if you want to add more explanation about custom kv cache here. We can update the measurement once have data on-device. Current number is what I got when running on host.
 
Screenshot 2025-06-05 at 5 54 27 PM

guangy10:
Tag @kimishpatel for review

README.md
```diff
@@ -107,9 +87,11 @@ optimum-cli export executorch \
     --model "HuggingFaceTB/SmolLM2-135M-Instruct" \
     --task "text-generation" \
     --recipe "xnnpack" \
-    --output_dir="hf_smollm2" \
     --use_custom_sdpa \
-    --qlinear
+    --use_custom_kv_cache \
+    --qlinear \
+    --qembedding \
```
guangy10, line 93: @kimishpatel Since we're adding all optimizations here, this path must be tested in CI, which we don't have yet. I will put another PR to cover this shortly

guangy10: Here: https://github.com/huggingface/optimum-executorch/pull/75

README.md
kimishpatel: I think this repo is aimed at quick and good out of box experience so I dont know if any customization is something we have spend too much time highlighting.

Also we should probably roll both custom sdpa and kv cache inside xnnpack recipe so users dont have to set any extra options

guangy10: I like the idea of hiding all these details inside xnnpack at some pony. I think it will improve the usability as users can get the optimal/default config out of the box. For now, I think it worth showcasing the composability/customization as we don’t know what would the optimal config for different types of models, even for xnnpack. For example, the optimizations for decoder-only may apply to encode-decoder models, the encoder part may need different quant config, so we may need to play with these different combinations.

PR #75, opened 2025-06-06
Cover custom sdpa + kv cache + quant in CI

guangy10:
As titled, showing all supported optimizations are composable and can be combined to achieve better perf.

### DIFF
--- a/tests/models/test_modeling_llama.py
+++ b/tests/models/test_modeling_llama.py
@@ -17,7 +17,6 @@ import gc
 import logging
 import os
 import subprocess
-import sys
 import tempfile
 import unittest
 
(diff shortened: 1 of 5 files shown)


guangy10:
@kimishpatel for review

tests/models/test_modeling_llama.py
```diff
@@ -128,3 +128,32 @@ def test_llama_text_generation_with_custom_sdpa_8da4w_8we(self):
...
+            prompt="Simply put, the theory of relativity states that",
+            max_seq_len=32,
+        )
+        logging.info(f"\nGenerated text:\n\t{generated_text}")
+        generated_tokens = tokenizer(generated_text, return_tensors="pt").input_ids
+
+        # Free memory before loading eager for quality check
+        del model
+        del tokenizer
+        gc.collect()
```
kimishpatel: you can factor out a bunch of this code in util so only one place needs update

kimishpatel (approved):
looks good but some refactor would be nice

tests/models/test_modeling_llama.py
guangy10: Yeah, I was thinking to parameterize these tests when I have time, but it’s not a priority, only the coverage in CI matters atm to ensure all pieces can work together. I can file a github task to refactor these tests

guangy10:
Will check the CI failures tomorrow

PR #77, opened 2025-06-06
Ensure custom sdpa only handles is_causal

kimishpatel:
Summary:
Custom sdpa when used with start_pos only works for causal attention

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:


guangy10 (approved):
Looks good

PR #78, opened 2025-06-06
    refactor: leverage ExecutorTorch's CustomKVCache for per-layer cache management

kimishpatel:
Summary:

        - Replace separate key/value cache tensors with CustomKVCache
          instances
        - Use CustomKVCache.update() method instead of manual
          torch.ops.llama.update_cache calls
        - Create self.kv_cache list with one CustomKVCache per layer
        - Maintain compatibility with parent StaticCache class
        - Fix type safety issues in get_seq_length method
        - Simplify cache update logic by leveraging ExecutorTorch's proven
          implementation

    Test Plan:
    CI

### DIFF
--- a/.github/workflows/test_models.yml
+++ b/.github/workflows/test_models.yml
@@ -51,18 +51,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies for ExecuTorch
         run: |
-          pip install '.[tests]'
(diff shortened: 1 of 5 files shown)


kimishpatel:
@guangy10 in prep for ring kv cache

kimishpatel:
@guangy10 rebased

guangy10 (approved):
LGTM

PR #79, opened 2025-06-09
Add CoreML recipe to optimum ExecuTorch

metascroy:
This PR adds a coreml recipe to optimum-executorch.

### DIFF
--- a/optimum/exporters/executorch/__main__.py
+++ b/optimum/exporters/executorch/__main__.py
@@ -130,6 +130,8 @@ def main_export(
     kwargs["force_download"] = force_download
     kwargs["config"] = config
 
+    recipe_kwargs = kwargs.pop("recipe_kwargs", {})
+
     model = task_func(model_name_or_path, **kwargs)
 
     if not os.path.exists(output_dir):
@@ -140,7 +142,7 @@ def main_export(
         task=task,
         recipe=recipe,
         output_dir=output_dir,
-        **kwargs,
+        **recipe_kwargs,
     )
 
 
--- a/optimum/exporters/executorch/recipes/coreml.py
+++ b/optimum/exporters/executorch/recipes/coreml.py
@@ -1,0 +1,154 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+from typing import Dict, Union
+
+from packaging.version import parse
+from tabulate import tabulate
+from torch.export import ExportedProgram
+import coremltools as ct
+import torch
+
+from executorch import version as executorch_version
+from executorch.backends.apple.coreml.partition import CoreMLPartitioner
+from executorch.backends.apple.coreml.compiler import CoreMLBackend
+
+from executorch.devtools.backend_debug import get_delegation_info
+from executorch.exir import (
+    EdgeCompileConfig,
+    ExecutorchBackendConfig,
+    ExecutorchProgram,
+    to_edge_transform_and_lower,
+)
+from optimum.executorch.passes.remove_padding_idx_embedding_pass import RemovePaddingIdxEmbeddingPass
+from executorch.backends.apple.coreml.quantizer import CoreMLQuantizer
+from torchao.quantization.pt2e.quantize_pt2e import convert_pt2e, prepare_pt2e
+from ..integrations import (
+    CausalLMExportableModule,
+    MaskedLMExportableModule,
+    Seq2SeqLMExportableModule,
+)
+from ..recipe_registry import register_recipe
+
+def get_quantization_config():
+    quantization_config = ct.optimize.torch.quantization.LinearQuantizerConfig.from_dict(
+        {
+            "global_config": {
+                "quantization_scheme": ct.optimize.torch.quantization.QuantizationScheme.symmetric,
+                "activation_dtype": torch.quint8,
+                "weight_dtype": torch.qint8,
+                "weight_per_channel": True,
+            }
+        }
+    )
+    return quantization_config
+
+def quantize_program(ep):
+    quantizer = CoreMLQuantizer(get_quantization_config())
+    gm = ep.module()
+    
+    args, kwargs = ep.example_inputs
+    prepared_model = prepare_pt2e(gm, quantizer) 
+    prepared_model(*args, **kwargs)
+    converted_model = convert_pt2e(prepared_model)
+    return torch.export.export(converted_model, args, kwargs)
+
+
+@register_recipe("coreml")
+def export_to_executorch_with_coreml(
+    model: Union[CausalLMExportableModule, MaskedLMExportableModule, Seq2SeqLMExportableModule],
+    **kwargs,
+):
+    """
+    Export a PyTorch model to ExecuTorch w/ delegation to CoreML backend.
+
+    This function also write metadata required by the ExecuTorch runtime to the model.
+
+    Args:
+        model (Union[CausalLMExportableModule, MaskedLMExportableModule, Seq2SeqLMExportableModule]):
+            The PyTorch model to be exported to ExecuTorch.
+        **kwargs:
+            Additional keyword arguments for recipe-specific configurations, e.g. export using different example inputs, or different compile/bechend configs.
+
+    Returns:
+        Dict[str, ExecutorchProgram]:
+            A map of exported and optimized program for ExecuTorch.
+            For encoder-decoder models or multimodal models, it may generate multiple programs.
+    """
+
+    def _lower_to_executorch(
+        exported_programs: Dict[str, ExportedProgram],
+        metadata=None,
+        **kwargs,
+    ) -> Dict[str, ExecutorchProgram]:
+
+        minimum_deployment_target = kwargs.get("minimum_ios_deployment_target", "15")
+        minimum_deployment_target = {
+            "15": ct.target.iOS15,
+            "16": ct.target.iOS16,
+            "17": ct.target.iOS17,
+            "18": ct.target.iOS18,
+        }[minimum_deployment_target]
+
+        compute_precision = kwargs.get("compute_precision", "fp16")
+        compute_precision = {
+            "fp16": ct.precision.FLOAT16,
+            "fp32": ct.precision.FLOAT32,
+        }[compute_precision]
+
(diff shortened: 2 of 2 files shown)


optimum/exporters/executorch/recipes/coreml.py
```diff
@@ -0,0 +1,154 @@
...
+                "quantization_scheme": ct.optimize.torch.quantization.QuantizationScheme.symmetric,
+                "activation_dtype": torch.quint8,
+                "weight_dtype": torch.qint8,
+                "weight_per_channel": True,
+            }
+        }
+    )
+    return quantization_config
+
+def quantize_program(ep):
```
metascroy: TODO: remove quantize_program

optimum/exporters/executorch/recipes/coreml.py
```diff
@@ -0,0 +1,154 @@
...
+            For encoder-decoder models or multimodal models, it may generate multiple programs.
+    """
+
+    def _lower_to_executorch(
+        exported_programs: Dict[str, ExportedProgram],
+        metadata=None,
+        **kwargs,
+    ) -> Dict[str, ExecutorchProgram]:
+
+        minimum_deployment_target = kwargs.get("minimum_ios_deployment_target", "15")
```
metascroy: Should the kwarg keys be things like ct.target.iOS15 directly, or is str/int "15" better?

optimum/exporters/executorch/recipes/coreml.py
```diff
@@ -0,0 +1,154 @@
...
+        minimum_deployment_target = {
+            "15": ct.target.iOS15,
+            "16": ct.target.iOS16,
+            "17": ct.target.iOS17,
+            "18": ct.target.iOS18,
+        }[minimum_deployment_target]
+
+        compute_precision = kwargs.get("compute_precision", "fp16")
```
metascroy: Should we use torch.float16 instead of "fp16" as key value?  Or maybe ct.precision.FLOAT16?

metascroy:
@guangy10 the tests are failing due to missing coremltools dependency.  Can I add coremltools to INSTALL_REQUIRE in setup.py?

optimum/exporters/executorch/__main__.py
```diff
@@ -140,7 +142,7 @@ def main_export(
         task=task,
         recipe=recipe,
         output_dir=output_dir,
-        **kwargs,
+        **recipe_kwargs,
```
guangy10, line 145: Currently we don't differentiate if the kwargs are for recipe or something else. We pass the entire kwargs to registered tasks and registered recipes and they look for the args that are relevant. The changes here may break the export.

optimum/exporters/executorch/recipes/coreml.py
```diff
@@ -0,0 +1,111 @@
...
+                f"\nExecuTorch program for {pte_name}.pte: {et_progs[pte_name].exported_program().graph_module}"
+            )
+            delegation_info = get_delegation_info(et_progs[pte_name].exported_program().graph_module)
+            logging.debug(f"\nDelegation info Summary for {pte_name}.pte: {delegation_info.get_summary()}")
+            logging.debug(
+                f"\nDelegation info for {pte_name}.pte: {tabulate(delegation_info.get_operator_delegation_dataframe(), headers='keys', tablefmt='fancy_grid')}"
+            )
+        return et_progs
+
+    exported_progs = model.export()
```
guangy10, line 115: Note that the input model is `Union[CausalLMExportableModule, MaskedLMExportableModule, Seq2SeqLMExportableModule]`, which has been wrapped in an exportable way, by taking the export recipe. Just a note that if the coreml requires export the model differently than XNNPACK, e.g. certain dims to be static/dynamic, you may want to take a look the export recipe

optimum/exporters/executorch/recipes/coreml.py
guangy10: I guess string is probably better if you want to pass it via CLI, i.e. via `optimum-cli export executorh`. If we have the similar arg in etLLM or export API, it's better to be consistent.

optimum/exporters/executorch/recipes/coreml.py
guangy10: HF API can take both the `torch.float16` or the string `"float16"` but not fp16, so I think it's better to pass "float16" if you'd like to pass this via the CLI for export (like what we are doing here: https://github.com/huggingface/optimum-executorch/blob/51f3f1feb4da0800c1b46ca06ab6eca4ba2a4839/tests/models/test_modeling_qwen3.py#L129-L135), or if want to hide this inside the coreml recipe, there is no difference between using torch.float16 and float16.

setup.py
```diff
@@ -15,6 +15,7 @@
     "optimum~=1.24",
     "executorch>=0.6.0",
     "transformers==4.51.3",
+    "coremltools==8.3.0",
```
guangy10: Maybe add it to `TESTS_ERQUIRE`? It is where the deps to run certain model/backends are added. We will probably refactor it later, but to be consistent for now, it's better adding it there.

guangy10 (changes requested):
@metascroy Sorry for taking long to get this PR reviewed. For the recipe, if it's not tailored for certain models, do you want to add tests to other models as well?

tests/models/test_modeling_efficientnet.py
```diff
@@ -76,3 +75,28 @@ def test_efficientnet_image_classification(self):
 
         # Compare with eager outputs
         self.assertTrue(check_close_recursively(eager_output.logits, et_output))
+    
+    # @slow
+    # @pytest.mark.run_slow
+    def test_efficientnet_image_classification_coreml(self):
```
guangy10: You may want to gate this test to run on macOS runner only. You can do something like this https://github.com/huggingface/optimum-executorch/blob/main/tests/models/test_modeling_gemma3.py#L113

optimum/exporters/executorch/recipes/coreml.py
metascroy: CoreML can handle static/dynamic dims.  I think the goal is for having the same export for both XNNPACK/CoreML delegation.

metascroy:
> @metascroy Sorry for taking long to get this PR reviewed. For the recipe, if it's not tailored for certain models, do you want to add tests to other models as well?

The recipe is not tailed for certain models.  I don't know the best way to add tests because all the current tests look like they have XNNPACK hardcoded, but it would be nice the model tests could be parametrized on delegation recipe or something.

optimum/exporters/executorch/__main__.py
metascroy: I first used kwargs for the recipe_kwargs, but then I got some errors like "task got unexpected keyword argument minimum_deployment_target" (which is a CoreML kwarg).  Are you sure you can use the same kwargs for both?

optimum/exporters/executorch/recipes/coreml.py
metascroy: > HF API can take both the `torch.float16` or the string `"float16"` but not fp16, so I think it's better to pass "float16" if you'd like to pass this via the CLI for export (like what we are doing here:
> 
> https://github.com/huggingface/optimum-executorch/blob/51f3f1feb4da0800c1b46ca06ab6eca4ba2a4839/tests/models/test_modeling_qwen3.py#L129-L135
> 
> ), or if want to hide this inside the coreml recipe, there is no difference between using torch.float16 and float16.

I ended up going with ct.PRECISION.FLOAT16 to match the other CoreML args.

guangy10:
> > @metascroy Sorry for taking long to get this PR reviewed. For the recipe, if it's not tailored for certain models, do you want to add tests to other models as well?
> 
> The recipe is not tailed for certain models. I don't know the best way to add tests because all the current tests look like they have XNNPACK hardcoded, but it would be nice the model tests could be parametrized on delegation recipe or something.
> 
> Thoughts?

There is a plan to parametrize the tests, but it's not a priority so nobody actually worked on it. The setup of this repo is taking the same principle of transformers does, single test single file. Duplicate test code is completely fine, the coverage and test integrity in the CI matters most.

optimum/exporters/executorch/__main__.py
guangy10: > I got some errors like "task got unexpected keyword argument minimum_deployment_target"

From which task?

> Are you sure you can use the same kwargs for both?

Here is the code pointer: https://github.com/huggingface/optimum-executorch/blob/main/optimum/exporters/executorch/__main__.py#L133-L144

metascroy: For example, compute_precision is a coreml recipe kwarg, but if I use the same kwargs for both the task and recipe, I'll get an error like:

"E           TypeError: ViTForImageClassification.__init__() got an unexpected keyword argument 'compute_precision'"

when running test_vit_image_classification.

Do you have an example of using the same kwargs for both tasks/recipes that works?  xnnpack doesn't look like it uses kwargs in the recipe.

guangy10: Oh I mean we're passing all `kwargs` down to both the registered recipes and regiestered tasks, it's up to each task/recipe to look up and use relevant args, or completely ignore `kwargs`.

optimum/exporters/executorch/recipes/coreml.py
guangy10: Can you add a test for export via optimum CLI?

optimum/exporters/executorch/__main__.py
metascroy: What is your suggestion here @guangy10?

 ViTForImageClassification is not a task I wrote.  It is an existing task, and it doesn't seem to like receiving a kwarg it doesn't expect.  So I don't see how the recipe and task can share kwargs without rewriting a lot of tasks.

guangy10: > ViTForImageClassification is not a task I wrote. It is an existing task, and it doesn't seem to like receiving a kwarg it doesn't expect. So I don't see how the recipe and task can share kwargs without rewriting a lot of tasks.

@metascroy I think you will only need to modify this line: https://github.com/huggingface/optimum-executorch/blob/main/optimum/exporters/executorch/tasks/image_classification.py#L41

I don't think we are setting any `kwargs` for vision models so for simplicity the code is just directly passing all `kwargs` to the `AutoModelForImageClassification`. This is where you need to extract and copy task relevant args and only pass those down to `AutoModelForImageClassification`

metascroy: @guangy10 I can make this change, but do you really think this is the best design?  It seems I’ll have to make a similar change for all task types.

What kwargs should I pass to the vision model?  An empty dict?

guangy10: @metascroy TBH I don't really see a concrete problem we're trying to solve here, so not sure when you say "best design" what it really means. I'm not suggesting it's the way to go I was just showing a way to unblock. I will be happy to take your approach as long as it's backwards compatible. Note that repurpose the top-level `kwargs` to be recipe only is not BC. 

For historical reason the code structure and some mechanism are implemented by following the same convention of other optimum backends when integrated to `optimum` repo. Note that executorch was one of the backends in the parent `optimum` repo and early this year it was separated to a dedicated repo, but the code structure doesn't really change. Something may or may not make sense given it's not longer in optimum. I was thinking to refactor it but it's not a priority.

PR #81, opened 2025-06-10
Add support for sliding window via ring kv cache

kimishpatel:
Summary:
This diffs adds
- Ring kv cache for local attention where ring buffer updates cache with new entries in a ring fashion rather than sliding cache
- Correspondingly we have to update sdpa to accept mask that correspond to the state of ring buffer because slot in ring buffer may correspond to different token positions. Thus we cannot just use causal mask.

Now custom_kv_cache option means we support both global attention and local/global like gemma3

Test Plan:
Test added

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/optimum/exporters/executorch/integrations.py
+++ b/optimum/exporters/executorch/integrations.py
@@ -26,6 +26,7 @@ from transformers import (
 )
 from transformers.generation.configuration_utils import GenerationConfig
 
+from optimum.executorch.attentions.custom_sdpa import get_custom_sdpa_for_ring_kv_cache
 from optimum.utils.import_utils import is_transformers_version
 
 from .utils import save_config_to_constant_methods
@@ -56,6 +57,12 @@ class CausalLMExportableModule(torch.nn.Module):
             max_batch_size = 1
             max_cache_len = 4094
             exportable_module = TorchExportableModuleForDecoderOnlyLM(self.model, max_batch_size, max_cache_len)
+
+            from transformers.modeling_utils import AttentionInterface
+
+            _custom_sdpa_for_ring_kv_cache = get_custom_sdpa_for_ring_kv_cache(exportable_module)
+            AttentionInterface.register("custom_sdpa_ring_kv_cache", _custom_sdpa_for_ring_kv_cache)
+            exportable_module.model.model.config._attn_implementation = "custom_sdpa_ring_kv_cache"
             if self.use_custom_kv_cache:
                 from optimum.executorch.attentions.custom_kv_cache import (
                     replace_with_et_custom_kv_cache,
--- a/optimum/exporters/executorch/integrations.py
+++ b/optimum/exporters/executorch/integrations.py
@@ -38,12 +38,32 @@ class CausalLMExportableModule(torch.nn.Module):
     This module ensures that the exported model is compatible with ExecuTorch.
     """
 
-    def __init__(self, model, use_custom_kv_cache=False):
+    def __init__(self, model, use_custom_kv_cache=False, use_custom_sdpa=False):
         super().__init__()
         self.model = model
         self.config = model.config
         self.use_custom_kv_cache = use_custom_kv_cache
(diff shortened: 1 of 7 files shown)


tests/models/test_modeling_gemma3.py
```diff
@@ -215,3 +216,41 @@ def test_gemma3_text_generation_with_custom_sdpa_8da4w_8we(self):
...
+        model_id = "unsloth/gemma-3-1b-it"
+        prompt = "Write a poem about a machine learning."
+
+        # ExecuTorch model + custom sdpa + 8da4w linear quantization + int8 embedding quantization
+        kwargs = {"qlinear": True, "qembedding": True}
+        model = ExecuTorchModelForCausalLM.from_pretrained(
+            model_id,
+            recipe="xnnpack",
+            attn_implementation="custom_sdpa",
+            use_custom_kv_cache=True,
```
guangy10, line 236: @kimishpatel Do you forget to add custom_sdpa?

guangy10:
Pretty much all gemma3 tests are failing export. Does this change require bumping `transformers`? I suspect if that's the reason of the failures in CI

kimishpatel:
> Pretty much all gemma3 tests are failing export. Does this change require bumping `transformers`? I suspect if that's the reason of the failures in CI

Let me try to repro the failure

optimum/exporters/executorch/integrations.py
```diff
@@ -48,14 +49,23 @@ def export(self, input_ids=None, cache_position=None) -> Dict[str, ExportedProgr
...
             max_batch_size = 1
             max_cache_len = 4094
             exportable_module = TorchExportableModuleForDecoderOnlyLM(self.model, max_batch_size, max_cache_len)
+
+            from transformers.modeling_utils import AttentionInterface
+
+            _custom_sdpa_for_ring_kv_cache = get_custom_sdpa_for_ring_kv_cache(exportable_module)
+            AttentionInterface.register("custom_sdpa_ring_kv_cache", _custom_sdpa_for_ring_kv_cache)
+            AttentionMaskInterface.register("custom_sdpa_ring_kv_cache", sdpa_mask_without_vmap)
+            exportable_module.model.model.config._attn_implementation = "custom_sdpa_ring_kv_cache"
```
kimishpatel: This might be the reason why it works

guangy10: @kimishpatel Maybe we should abort https://github.com/huggingface/optimum-executorch/pull/86 into this PR so this PR will be testable in the CI?

guangy10: I don't quite follow why would this line "fixes" the quality issue in other models like smollm that are not using ring_kv_cache. Do you mind elaborate a bit?

guangy10: I know that in Cyril's vision the static cache will be a special case of hybrid cache w/o sliding window cache then all models can technically work with hybrid cache then. Is it something you are referring here?

guangy10: @kimishpatel Can you absorb https://github.com/huggingface/optimum-executorch/pull/86 into this PR?

kimishpatel: > I don't quite follow why would this line "fixes" the quality issue in other models like smollm that are not using ring_kv_cache. Do you mind elaborate a bit?

Mainly because it is not gemma3/local attention specific. custom_sdpa_ring_kv_cache handles both global and sliding window attention

kimishpatel: > I know that in Cyril's vision the static cache will be a special case of hybrid cache w/o sliding window cache then all models can technically work with hybrid cache then. Is it something you are referring here?

yeah kind of

guangy10:
All look good. @kimishpatel We can get it merged if no more changes needed

kimishpatel:
there was one change that i had to push to make it work when attn_implementation is used for custom_sdpa

PR #82, opened 2025-06-15
Changes to accommodate new transformer changes around attention mask

kimishpatel:
Summary:
Post 4.53 custom attention must have attention mask function registered else mask will not be created.

Test Plan:
CI test on top of 4.53 for custom_sdpa although it requires upstream change in transformers

Reviewers:

Subscribers:

Tasks:

Tags:


kimishpatel:
@guangy10

guangy10 (approved):

PR #89, opened 2025-06-20
Update pinned executorch nightly

guangy10:
ExecuTorch nightly is unblocked. We can update the pin to pick up fixes that are required by prefill.
Synced all torch deps to the nightly from same day

### DIFF
--- a/install_dev.py
+++ b/install_dev.py
@@ -5,18 +5,21 @@ import sys
 
 def install_torch_nightly_deps():
     """Install torch related dependencies from pinned nightly"""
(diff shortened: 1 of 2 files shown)


guangy10:
@kimishpatel for review

guangy10:
Can't repro the issue locally with the updated ET pin. Checking other installed packages...

guangy10:
Fixed torchvision

guangy10:
hmm, the complained version `torch==2.8.0.dev20250601` can be installed without any problem on linux devserver

PR #90, opened 2025-06-24
Add resnet50 to supported models

metascroy:
This adds microsoft/resnet50 to support ET models, and adds a test for XNNPACK/CoreML lowering.

microsoft/resnet50 does not contain image_size in its config, so we introduce model_to_pixel_values_size to explicitly track example input sizes for models by id.


guangy10:
@metascroy Can you rebase before merging?

PR #95, opened 2025-06-26
Update recipe for Phi4-mini

guangy10:


guangy10:
@kimishpatel for review

optimum/exporters/executorch/tasks/causal_lm.py
```diff
@@ -60,7 +60,13 @@ def load_causal_lm_model(model_name_or_path: str, **kwargs) -> CausalLMExportabl
     cache_implementation = kwargs.get("cache_implementation", "static")
     use_custom_sdpa = use_custom_sdpa or attn_implementation == "custom_sdpa"
     max_length = kwargs.get("max_length", 2048)
-    config = kwargs.get("config", None)
+    config = kwargs.get("config") or AutoConfig.from_pretrained(model_name_or_path)
+
+    if hasattr(config, "rope_scaling") and config.rope_scaling is not None:
```
kimishpatel, line 65: why do we need such changes?

tests/models/test_modeling_phi4.py
```diff
@@ -40,41 +44,46 @@ def __init__(self, *args, **kwargs):
...
     @pytest.mark.skipif(
-        is_ci,
-        reason="Test Phi-4-mini (3.8B) will require runner to be configured with larger RAM",
+        is_linux_ci
+        or parse(transformers.__version__) < parse("4.52.0")
+        or parse(torchao.__version__) < parse("0.11.0"),
+        reason="Only available on transformers >= 4.52.0 and torchao >= 0.11.0. OOM on linux runner.",
     )
-    def test_phi4_text_generation(self):
+    def test_phi4_text_generation_with_custom_sdpa_and_kv_cache_8da4w_8we(self):
```
kimishpatel, line 52: should you not make a separate test?

optimum/exporters/executorch/tasks/causal_lm.py
guangy10: Yeah, config object can be passed via from_pretrained but can't via optimum-cli. For phi-4 particular, the default config is using long rope which has data-dependent code that causes export issue, we need to override that config to use the default rope instead. Because there is no easy way to pass it via optimum-cli (unless we pass the entire json), it's better we always load the default config here no matter the export workflow is triggered via optimum-cli or from_pretrained.

tests/models/test_modeling_phi4.py
guangy10: This model is large (3.2B IIUC), the previous recipe for Phi-4 is not running in CI due to OOM, so it doesn't matter.

optimum/exporters/executorch/tasks/causal_lm.py
kimishpatel: I just want to minimize the customization that we have to do here. Ideally I want to structure it such that this repo can be simplified to a bunch of scripts

PR #97, opened 2025-07-04
Add link to ET's tutorial to run the PTE on-device

guangy10:
Since we linked the benchmark table/numbers, we want to also provide instructions for users to be able to repro and run on their own device. Instead of duplicate the doc in optimum-et, we should just redirect users to follow the ET's instruction page to build and deploy to the runner or apps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ optimum-cli export executorch \
 ```
 Explore the various export options by running the command: `optimum-cli export executorch --help`
 
-#### Step 2: Load and run inference
+#### Step 2: Validate the Exported Model on Host Using the Python API
 Use the exported model for text generation:
 ```python
 from optimum.executorch import ExecuTorchModelForCausalLM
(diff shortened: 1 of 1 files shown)


jackzhxng (approved):
Maybe https://github.com/pytorch/executorch/blob/main/examples/models/llama/README.md#step-5-build-mobile-apps?

guangy10:
> Embed a screenshot or something from ET repo

The screenshot of the instructions? I'd rather leave it text-only atm as the linked instructions are for Qwen3. We can update it once we have a more generic tutorial from ET

README.md
```diff
@@ -115,6 +115,9 @@ generated_text = model.text_generation(
 print(generated_text)
 ```
 
+#### Step 3: Run inference on-device
+To perform on-device inference, you can use ExecuTorch’s sample runner or the example iOS/Android applications. For detailed instructions, refer to the [ExecuTorch Sample Runner guide](https://github.com/pytorch/executorch/blob/main/examples/models/qwen3/README.md#example-run).
```
guangy10, line 119: The linked instructions is for Qwen3 and in ET's repo these instructions are duplicated for phi4/llama, which doesn't look optimal to me. @jackzhxng As part of the doc revamp work, I think we should first consolidate the readme with more generic instructions, and then update the link here. I'll leave this comment on as a TODO or we can file a GitHub Issue to track it

guangy10:
> Maybe https://github.com/pytorch/executorch/blob/main/examples/models/llama/README.md#step-5-build-mobile-apps?

That doesn't cover run with llama_runner

PR #101, opened 2025-07-10
Add portable recipe and tests

manuelcandales:
This PR introduced a new portable recipe, intended to test model coverage of ExecuTorch portable kernels.
Tests have been added for each currently supported model. A portable marker has been incorporated into the pytest configuration within pyproject.toml, enabling the selection and running of portable tests only.

### DIFF
--- a/optimum/exporters/executorch/recipes/portable.py
+++ b/optimum/exporters/executorch/recipes/portable.py
@@ -1,0 +1,94 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+from typing import Dict, Union
+
+from torch.export import ExportedProgram
+
+from executorch.exir import (
+    EdgeCompileConfig,
+    ExecutorchProgram,
+    to_edge_transform_and_lower,
+)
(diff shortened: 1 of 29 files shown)


optimum/exporters/executorch/recipes/portable.py
```diff
@@ -0,0 +1,94 @@
...
+    CausalLMExportableModule,
+    MaskedLMExportableModule,
+    Seq2SeqLMExportableModule,
+)
+from ..recipe_registry import register_recipe
+
+
+@register_recipe("portable")
+def export_to_executorch_with_portable(
+    model: Union[CausalLMExportableModule, MaskedLMExportableModule, Seq2SeqLMExportableModule],
```
guangy10, line 37: The model is loaded and wrapped in the export recipe. Should the portable recipe also cover dtypes float32, float16 and bfloat16?

manuelcandales: I thought dtype was controlled via kwargs?

guangy10: Yeah, but I want to double check if the kernels gaps you are trying to figure out through this exercise would cover all dtypes mentioned

optimum/exporters/executorch/recipes/portable.py
```diff
@@ -0,0 +1,94 @@
...
+        et_progs = {}
+
+        for pte_name, exported_program in exported_programs.items():
+            logging.debug(f"\nExported program for {pte_name}.pte: {exported_program}")
+            et_progs[pte_name] = to_edge_transform_and_lower(
+                exported_program,
+                partitioner=[],
+                compile_config=EdgeCompileConfig(
+                    _check_ir_validity=False,
+                    _skip_dim_order=True,
```
digantdesai, line 70: Comment why? Or what would happen if you enable this?

optimum/exporters/executorch/recipes/portable.py
manuelcandales: I am only covering fp32 during spring 1. float16 and bfloat16 will be tested later.

optimum/exporters/executorch/recipes/portable.py
manuelcandales: I imagine some tests might fail if we enable this at this point, although I haven't tested, since the focus is not on dim order for this sprint. By the way, this is the same `EdgeCompileConfig` of the xnnpack recipe.

PR #102, opened 2025-07-11
Update CoreML recipe and add LLM test

metascroy:
This updates CoreML recipes to be export friendly, and adds test for SmolLM.

It also adds option to disable dynamic shapes

### DIFF
--- a/install_dev.py
+++ b/install_dev.py
@@ -5,7 +5,7 @@ import sys
 
 def install_torch_nightly_deps():
     """Install torch related dependencies from pinned nightly"""
-    EXECUTORCH_NIGHTLY_VERSION = "dev20250625"
+    EXECUTORCH_NIGHTLY_VERSION = "dev20250629"
     TORCHAO_NIGHTLY_VERSION = "dev20250620"
     # Torch nightly is aligned with pinned nightly in https://github.com/pytorch/executorch/blob/main/install_requirements.py#L74
     TORCH_NIGHTLY_VERSION = "dev20250601"
--- a/optimum/exporters/executorch/recipes/coreml.py
+++ b/optimum/exporters/executorch/recipes/coreml.py
@@ -34,8 +34,7 @@ from ..integrations import (
 from ..recipe_registry import register_recipe
 
 
-@register_recipe("coreml")
-def export_to_executorch_with_coreml(
+def _export_to_executorch(
     model: Union[CausalLMExportableModule, MaskedLMExportableModule, Seq2SeqLMExportableModule],
     **kwargs,
 ):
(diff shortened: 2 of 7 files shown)


install_dev.py
```diff
@@ -5,7 +5,7 @@
 
 def install_torch_nightly_deps():
     """Install torch related dependencies from pinned nightly"""
-    EXECUTORCH_NIGHTLY_VERSION = "dev20250625"
+    EXECUTORCH_NIGHTLY_VERSION = "dev20250627"
```
guangy10, line 8: Just curious, what are the new features/fixes from ET?

install_dev.py
```diff
@@ -5,7 +5,7 @@
...
     TORCHAO_NIGHTLY_VERSION = "dev20250620"
     # Torch nightly is aligned with pinned nightly in https://github.com/pytorch/executorch/blob/main/install_requirements.py#L74
     TORCH_NIGHTLY_VERSION = "dev20250601"
```
guangy10, line 11: Should update the pinned PT accordingly. Source of truth is the one pinned on ET trunk. To avoid testing multiple version combinations of different libraries

optimum/exporters/executorch/recipes/coreml.py
```diff
@@ -85,18 +101,24 @@ def _lower_to_executorch(
...
                             model_type=model_type,
+                            op_linear_quantizer_config=op_linear_quantizer_config,
                         ),
                         take_over_mutable_buffer=take_over_mutable_buffer,
                     )
                 ],
                 compile_config=EdgeCompileConfig(
                     _check_ir_validity=False,
-                    _skip_dim_order=False,
+                    # In ET 0.7, we can set _skip_dim_order=False
```
guangy10, line 120: Are you saying there is a regression in the pinned ET? 0.7 has been cut already, new nightly should be created for 0.7. Do we still need to toggle this on/off?

optimum/exporters/executorch/tasks/causal_lm.py
```diff
@@ -61,6 +61,7 @@ def load_causal_lm_model(model_name_or_path: str, **kwargs) -> CausalLMExportabl
     use_custom_sdpa = use_custom_sdpa or attn_implementation == "custom_sdpa"
     max_length = kwargs.get("max_length", 2048)
     config = kwargs.get("config") or AutoConfig.from_pretrained(model_name_or_path)
+    disable_dynamic_shapes = kwargs.get("disable_dynamic_shapes", False)
```
guangy10, line 64: nit: Maybe change the name to `use_dynamic_shapes` and default to `True` as it will be the mainstream for both XNNPACK and CoreML.

guangy10 (approved):
Looks good to me. Need to fix the CI before merging

PR #103, opened 2025-07-11
Update ET pin

metascroy:


install_dev.py
```diff
@@ -5,7 +5,7 @@
 
 def install_torch_nightly_deps():
     """Install torch related dependencies from pinned nightly"""
-    EXECUTORCH_NIGHTLY_VERSION = "dev20250625"
+    EXECUTORCH_NIGHTLY_VERSION = "dev20250627"
     TORCHAO_NIGHTLY_VERSION = "dev20250620"
     # Torch nightly is aligned with pinned nightly in https://github.com/pytorch/executorch/blob/main/install_requirements.py#L74
     TORCH_NIGHTLY_VERSION = "dev20250601"
```
guangy10: For nightly, we should update the pinned torch nightly as well https://github.com/pytorch/executorch/blob/main/install_requirements.py#L74. That's the only combination of various versions we'd like to cover in the CI

metascroy: But this looks like the torch nightly for the pinned commit?

metascroy: Also, it’s mostly the macOS jobs that are failing.  The Linux all pass.

guangy10: > /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
  
That typically means OOM or something on the runner. Let's retry the jobs first.

guangy10: @metascroy The runners are fine. If we don't change the pinned ET, the nightly tests are working https://github.com/huggingface/optimum-executorch/pull/104. I think something is off in ET.

PR #106, opened 2025-07-17
Update et pin (attempt 2)

metascroy:


metascroy:
I don't see these issues locally @guangy10, e.g., I ran the efficientnet test on this PR and it passes.

Any ideas here?

PR #108, opened 2025-07-22
Coreml recipe updates

metascroy:
Update CoreML recipe to be CLI friendly.


guangy10:
Failures are irrelevant. I will create a GitHub issue for address the failure separately.

PR #112, opened 2025-07-24
Jz/voxtral

jackzhxng:


optimum/executorch/attentions/custom_sdpa.py
```diff
@@ -45,7 +45,7 @@ def custom_sdpa_with_start_pos_forward(
 
     # Ignore the causal flag from kwargs but use the one in module
     kwargs.pop("is_causal", None)
-    assert module.is_causal, "Current variant supports only causal attention"
+    # assert module.is_causal, "Current variant supports only causal attention"
```
larryliu0820, line 48: I don't know if this supports non causal case

PR #115, opened 2025-07-29
Refactor quantization and add quantization options to masked modeling

metascroy:
This PR refactors quantization into a common location, and adds quantization support to masked LM models (previously only available in decoder models).


metascroy:
cc @michaelbenayoun @echarlaix @guangy10

guangy10:
LGTM. Thanks for the fix

guangy10 (approved):

metascroy:
I think the failing test is not related.  Let me know if this can be merged @michaelbenayoun @echarlaix @guangy10

PR #117, opened 2025-08-01
Bump transformers to 4.54.1

jackzhxng:
Includes changes to absorb the https://github.com/huggingface/transformers/pull/39106 kv cache refactor.

PR #118, opened 2025-08-01
Bump transformers pin

larryliu0820:
TSIA

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ except Exception as error:
 INSTALL_REQUIRE = [
     "optimum~=1.24",
     "executorch>=0.6.0",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -14,7 +14,7 @@
 INSTALL_REQUIRE = [
     "optimum~=1.24",
     "executorch>=0.6.0",
-    "transformers==4.51.3",
+    "transformers==4.53.2",
```
guangy10, line 17: et 0.6.0 may not work with `transformers==4.53.2`.

larryliu0820: close in favor of https://github.com/huggingface/optimum-executorch/pull/117

PR #126, opened 2025-08-19
Add Voxtral

jackzhxng:
### Summary
Add `ExecuTorchModelForMultiModalToText` to handle early fusion vision and audio text-to-text models.

Note - cannot merge until 4.54 is released on `transformers` to pull in upstreamed changes

### Results
```
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████████████████████| 2/2 [00:00<00:00,  2.11it/s]
I tokenizers:regex.cpp:27] Registering override fallback regex
[program.cpp:134] InternalConsistency verification requested but not available
The audio features a conversation between two individuals, likely friends or acquaintances, who are engaged in a playful banter about tattoos. They are asking each other what their tattoos say, and the responses are all variations of the word "sweet." The conversation is humorous and lighthearted, with the individuals taking
⚠️ DISCLAIMER: Python-based perf measurements are approximate and may not match absolute speeds on Android/iOS apps. They are intended for relative comparisons—-e.g. SDPA vs. custom SDPA, FP16 vs. FP32—-so you can gauge performance improvements from each optimization steep. For end-to-end, platform-accurate benchmarks, please use the official ExecuTorch apps:

PyTorchObserver {"prompt_tokens": 1138, "generated_tokens": 63, "model_load_start_ms": 0, "model_load_end_ms": 0, "inference_start_ms": 1755652082353, "token_encode_end_ms": 1755652083721, "model_execution_start_ms": 0, "model_execution_end_ms": 0, "inference_end_ms": 1755652110934, "prompt_eval_end_ms": 1755652102257, "first_token_ms": 1755652102416, "aggregate_sampling_time_ms": 27193, "SCALING_FACTOR_UNITS_PER_SECOND": 1000}
        Prompt Tokens: 1138 Generated Tokens: 63
        Total inference time:           28.581000 (seconds)              Rate:  2.204262 (tokens/second)
                Prompt evaluation:      19.904000 (seconds)              Rate:  57.174437 (tokens/second)
                Generated 63 tokens:    8.677000 (seconds)               Rate:  7.260574 (tokens/second)
        Time to first generated token:  20.063000 (seconds)
        Sampling time over 1201 tokens: 27.193000 (seconds)
```

Basic stats
- PTE size: 2.7G
- Memory: 5.4 GB (this is just from `top`, not sure how much is this from other things such as auto processor that hasn't been garbage collected)


mergennachin:
> Add ExecuTorchModelForMultiModalToText

Will it work for Gemma3 too? Shall we add a unit test to test the vision-text-input using the new class?

optimum/executorch/modeling.py
```diff
@@ -1139,3 +1139,215 @@ def transcribe(
...
+                raise ValueError(
+                    f"Exported .pte file needs to contain the following required methods: {required_methods}"
+                )
+
+        self.encoder_name = None
+        for method_name in self.model.method_names():
+            if method_name == "audio_encoder":
+                self.encoder_name = "audio_encoder"
+            elif method_name == "vision_encoder":
+                self.encoder_name = "vision_encoder"
```
mergennachin: why is it mutually exclusive? For instance, gemma3n expects simultaneous audio and vision

mergennachin: If a model has both audio and vision encoders, only vision will work (last assignment wins).

optimum/executorch/modeling.py
```diff
@@ -1139,3 +1139,215 @@ def transcribe(
...
+
+    def forward(
+        self,
+        input_ids: torch.Tensor,
+        cache_position: torch.Tensor,
+        input_features: Optional[torch.Tensor] = None,
+    ):
+        token_embeddings = self.model.run_method("token_embeddings", (input_ids,))[0]
+        if input_features is not None:
+            token_embeddings = self.model.run_method(
```
mergennachin: Add validation that encoder output shape matches decoder input expectations

mergennachin: handle error if encoder fails

optimum/executorch/modeling.py
```diff
@@ -1139,3 +1139,215 @@ def transcribe(
...
+        first_token_generated = False
+        while len(generated_tokens) + prompt_tokens.size(1) < max_seq_len:
+            self.stats.on_sampling_begin()
+            logits = self.forward(
+                input_ids=torch.tensor([next_token], dtype=torch.long, device=self.device).unsqueeze(0),
+                cache_position=torch.tensor(
+                    [pos_base + len(generated_tokens) + prompt_tokens.size(1) - 1],
+                    dtype=torch.long,
+                    device=self.device,
+                ),
```
mergennachin, line 1281: explicitly set `input_features=None`

optimum/executorch/modeling.py
```diff
@@ -1139,3 +1139,215 @@ def transcribe(
...
+                Will be truncated to maximal cache size if larger than `max_cache_size`.
+        """
+        self.tokenizer = tokenizer
+
+        # Sanity check
+        if self.tokenizer.bos_token_id is not None and self.tokenizer.bos_token_id != self.bos_token_id:
+            raise ValueError(
+                f"The tokenizer's bos_token_id={self.tokenizer.bos_token_id} must be the same as the model's bos_token_id={self.bos_token_id}."
+            )
+        if isinstance(self.tokenizer, PreTrainedTokenizer) and not verify_eos_tokens_in_pretrained_tokenizer(
```
mergennachin, line 1329: what about supporting multiple eos_token_ids as opposed to single token

optimum/executorch/modeling.py
```diff
@@ -1139,3 +1139,215 @@ def transcribe(
...
+                first_token_generated = True
+
+            next_token = torch.argmax(logits[:, -1, :], dim=-1).item()
+            generated_tokens.append(next_token)
+            print(self.tokenizer.decode([next_token]), end="")
+
+            if next_token == self.eos_token_id:
+                break
+
+        self.stats.set_num_generated_tokens(len(generated_tokens) - len(prompt_tokens))
```
mergennachin: Isn't it prompt_tokens.size(1)?

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,291 @@
...
+from transformers.integrations.executorch import TorchExportableModuleForDecoderOnlyLM, sdpa_mask_without_vmap
+from transformers.masking_utils import AttentionMaskInterface
+from transformers.modeling_utils import AttentionInterface
 
 from optimum.executorch.attentions.custom_sdpa import get_custom_sdpa_for_ring_kv_cache
 
 from .utils import save_config_to_constant_methods
 
 
+class VoxtralEncoderExportableModule(torch.nn.Module):
```
mergennachin, line 40: why is this model specific implementation in integrations.py?

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,291 @@
...
+    """
+
+    def __init__(
+        self,
+        model: torch.nn.Module,
+        modality: str,
+        encoder_name: str,
+        use_custom_kv_cache: bool = False,
+        use_custom_sdpa: bool = False,
+    ):
```
mergennachin, line 122: can you do a few validations? 

encoder_name exists on the model,  check that model.config.text_config exists...

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,291 @@
...
+
+            input_ids = torch.zeros_like(inputs_embeds[:, :, 0], dtype=torch.long)
+            input_ids[0, 1] = (
+                self.config.audio_token_id if self.modality == "audio" else self.config.image_token_id
+            )  # Make sure we don't have an all-false mask for the imput_embeds.
+
+            if isinstance(self.model, VoxtralForConditionalGeneration):
+                encoder = VoxtralEncoderExportableModule(self.model)
+                input_features, dynamic_shapes = encoder.prepare_export_inputs()
+            else:
```
mergennachin, line 293: Why hardcode?

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,291 @@
...
+            if self.use_custom_sdpa:
+                getattr(self.model, self.encoder_name).config._attn_implementation = "custom_sdpa"
+            else:
+                getattr(self.model, self.encoder_name).config._attn_implementation = "sdpa_without_vmap"
+
+            input_ids = torch.zeros_like(inputs_embeds[:, :, 0], dtype=torch.long)
+            input_ids[0, 1] = (
+                self.config.audio_token_id if self.modality == "audio" else self.config.image_token_id
+            )  # Make sure we don't have an all-false mask for the imput_embeds.
+
```
mergennachin: - Hardcodes position [0, 1] for special token. Not all models put special token at position 1?
- No validation: Doesn't check if audio_token_id exists

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+import torchao
+from transformers import AutoConfig, AutoModel, GenerationConfig
+
+from ..integrations import MultiModalTextToTextExportableModule
+from ..quantization import quantize_model_
+from ..task_registry import register_task
+
+
```
mergennachin, line 26: we should open door for audio-vision-text input mode

mergennachin (changes requested):

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
```diff
@@ -0,0 +1,179 @@
...
+    return decoder_name, audio_encoder_name, vision_encoder_name
+
+
+# NOTE: It's important to map the registered task name to the pipeline name in https://github.com/huggingface/transformers/blob/main/utils/update_metadata.py.
+# This will streamline using inferred task names and make exporting models to Hugging Face pipelines easier.
+@register_task("image-text-to-text")
+@register_task("audio-text-to-text")
+@register_task("multimodal-text-to-text")  # Fake task name, since audio-text-to-text is not available.
+def load_multimodal_text_to_text_model(model_name_or_path: str, **kwargs):
```
kimishpatel, line 92: what is the right task that transformer uses?

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
```diff
@@ -0,0 +1,179 @@
...
+        "text_model",
+    ]
+    POTENTIAL_AUDIO_ENCODER_NAMES = [
+        "audio_tower",
+        "audio_model",
+    ]
+    POTENTIAL_VISION_ENCODER_NAMES = [
+        "vision_tower",
+        "vision_model",
+    ]
```
kimishpatel, line 48: We need to talk with hf on this. Else we continue to be at their mercy.

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
kimishpatel: i see that it is audio-text-to-text

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
```diff
@@ -0,0 +1,179 @@
...
+                - max_length (int, optional):
+                    Maximum sequence length for generation (default: 2048).
+
+    Returns:
+        MultiModalTextToTextExportableModule:
+            An instance of `MultiModalTextToTextExportableModule` for exporting and lowering to ExecuTorch.
+    """
+    device = "cpu"
+    batch_size = 1
+    dtype = kwargs.get("dtype", "float32")
```
kimishpatel, line 119: is dtype really somethng we allow to override? use_custom_sdpa doesnt work with fp16

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
```diff
@@ -0,0 +1,179 @@
...
+    config = kwargs.get("config") or AutoConfig.from_pretrained(model_name_or_path)
+
+    # Make sure config has text_config and vision_config:
+    if not (hasattr(config, "text_config") and (hasattr(config, "vision_config") or hasattr(config, "audio_config"))):
+        raise ValueError(
+            f"The model {model_name_or_path} does not have a `text_config` or `vision_config`/`audio_config` attribute in its config. "
+            "This is required for multimodal text-to-text models."
+        )
+
+    if hasattr(config, "rope_scaling") and config.rope_scaling is not None:
```
kimishpatel, line 156: why this? other tasks done have this

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
```diff
@@ -0,0 +1,179 @@
...
+        # NOTE: Avoid hitting the data-dependent control flow in _longrope_frequency_update.
+        config.rope_scaling["type"] = "default"
+    if hasattr(config, "use_cache") and config.use_cache is False:
```
kimishpatel, line 159: is it every not true?

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
```diff
@@ -0,0 +1,179 @@
...
+                "batch_size": batch_size,
+                "max_cache_len": max_length,
+            },
+        ),
+    )
+    decoder_name, audio_encoder_name, vision_encoder_name = _validate_multimodal_components(eager_model)
+    # Need to do this since apparently when nested modules (e.g. model.language_model) access the .property
+    # config, it always comes from the generation_config.json file, not the `generation_config` override
+    # from from_pretrained().
+    getattr(eager_model, decoder_name).generation_config = eager_model.generation_config
```
kimishpatel, line 182: definitely strange. do you wanna tag someone from transformer to get clarification on this?

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
```diff
@@ -0,0 +1,179 @@
...
+
+    for param in eager_model.parameters():
+        # Must disable gradient for quantized checkpoint
+        if isinstance(param, torchao.utils.TorchAOBaseTensor):
+            param.requires_grad = False
```
kimishpatel: nit: probably can be refactored to some utils. i see other tasks doing similar stuff

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+    The result of this subgraph should stream directly into the decoder subgraph.
+    """
+
+    def __init__(self, model: torch.nn.Module):
+        super().__init__()
+        self.audio_encoder = model.audio_tower
+        self.mm_projector = model.multi_modal_projector
+        self.intermediate_size = model.config.audio_config.intermediate_size
+        self.audio_token_id = model.config.audio_token_id
+        self.metadata = save_config_to_constant_methods(model.config.text_config, model.generation_config)
```
kimishpatel, line 52: is there anything about audio or vision config that need to be stored as well?

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+
+    def prepare_export_inputs(self):
+        chunk_length = (
+            self.audio_encoder.config.max_source_positions
+            * self.audio_encoder.conv1.stride[0]
+            * self.audio_encoder.conv2.stride[0]
```
kimishpatel, line 60: what are these for? is this a reliable way to get it? Looks like we are exploiting the fact that audio_encoder always has conv1/conv2.

Do add comment describing what this done.

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+        )
+        input_features = torch.rand(3, 128, chunk_length)  # (bsz, features, seq_len)
```
kimishpatel: where do these values come from? I think this is specific per model. Like voxtral has its own preprocessor_config.json. Is this not avaialbe via some config?

kimishpatel: or this 128 sounds like num_mel_bins in the config

optimum/exporters/executorch/integrations.py
kimishpatel: I dont follow why strides are being multiplied

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+
+        max_audio_len = 120  # In s, should be a multiple of 30. TODO: make configurable.
+        max_seq_len = self.metadata.get("get_max_seq_len")
+        dynamic_shapes = {
+            "input_features": {
+                0: torch.export.Dim("enc_batch_size_dim", min=1, max=max_audio_len // 30),
+            },
+        }
```
kimishpatel, line 69: i would have thought that the audio length corresponds to the last dim. Usually batch size does not include audio length. audio len should likely correspond to the chunk length.

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+
+    def prepare_export_inputs(self):
+        chunk_length = (
+            self.audio_encoder.config.max_source_positions
+            * self.audio_encoder.conv1.stride[0]
+            * self.audio_encoder.conv2.stride[0]
+        )
+        input_features = torch.rand(3, 128, chunk_length)  # (bsz, features, seq_len)
+
+        max_audio_len = 120  # In s, should be a multiple of 30. TODO: make configurable.
```
kimishpatel: whats the relation to multipl of 30?

optimum/exporters/executorch/integrations.py
kimishpatel: by default the strides are always 1 btw. I feel something is off here

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+            input_features (torch.FloatTensor): Raw audio features with shape (batch_size, features, seq_len).
+
+        Returns:
+            torch.FloatTensor: Combined embeddings with audio tokens replaced by audio embeddings.
+        """
+        audio_outputs = self.audio_encoder(input_features)
+        audio_hidden_states = audio_outputs.last_hidden_state
+        audio_hidden_states = audio_hidden_states.reshape(-1, self.intermediate_size)
+        audio_embeds = self.mm_projector(audio_hidden_states)
+        return audio_embeds
```
kimishpatel: So why do we need this https://github.com/huggingface/transformers/blob/main/src/transformers/models/voxtral/modeling_voxtral.py#L449-L453 here? Can the original hf model not be used directly?

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+        modality (str): The input modality type ("audio" or "vision").
+        encoder_name (str): Name of the encoder attribute in the model.
+        use_custom_kv_cache (bool): Whether to use custom key-value caching for optimization.
+        use_custom_sdpa (bool): Whether to use custom scaled dot-product attention.
+    """
+
+    def __init__(
+        self,
+        model: torch.nn.Module,
+        modality: str,
```
kimishpatel, line 116: we should definitely look into making it a more generic than only on modality as Mergen said

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+            input_ids (torch.Tensor): Example input IDs tensor.
+            dynamic_shapes (dict or None): Dynamic shape specifications for export.
+            strict (bool): Whether to use strict export mode.
+        """
+        seq_length = 3  # Sequence length > 1 to avoid specialization issues
+        example_input_ids = torch.zeros((1, seq_length), dtype=torch.long)
+
+        max_seq_len = self.metadata.get("get_max_seq_len")
+        sliding_window = self.metadata.get("sliding_window", float("inf"))
+        max_dim = min(max_seq_len, sliding_window) - 1
```
kimishpatel: put a loggin here when max_seq_len > sliding_window

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+        """
+
+        # Prepare inputs with dynamic shapes
+        seq_length = 3
+        example_inputs_embeds = torch.zeros((1, seq_length, self.config.text_config.hidden_size), dtype=torch.float)
+        example_cache_position = torch.arange(seq_length, dtype=torch.long)
+
+        max_seq_len = self.metadata.get("get_max_seq_len")
+        sliding_window = self.metadata.get("sliding_window", float("inf"))
+        max_dim = min(max_seq_len, sliding_window) - 1
```
kimishpatel: might be worth factoring this out

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+                from optimum.executorch.attentions.custom_kv_cache import (
+                    replace_with_et_custom_kv_cache,
+                )
+
+                replace_with_et_custom_kv_cache(
+                    exportable_module.model,
+                    self.model.config.text_config,
+                    self.model.generation_config,
+                    self.model.dtype,
+                )
```
kimishpatel: nit: also worth factoring out

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+            exported_programs["decoder"] = exported_program
+
+            # 2. Export token embeddings
+            input_ids, dynamic_shapes = self._prepare_text_embedding_export_inputs()
+            logging.info(
+                f"Exporting token embeddings using input_ids({input_ids.shape}), dynamic_shapes={dynamic_shapes}"
+            )
+
+            token_embeddings_exported_program = torch.export.export(
+                self.model.language_model.get_input_embeddings(),
```
kimishpatel: so language model export does not take care of embedding because we mix embedding for vision/audio part with language part?

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+                dynamic_shapes=dynamic_shapes,
+                strict=True,
+            )
+            exported_programs["token_embeddings"] = token_embeddings_exported_program
+
+            # 3. Export encoder.
+            if self.use_custom_sdpa:
+                getattr(self.model, self.encoder_name).config._attn_implementation = "custom_sdpa"
+            else:
+                getattr(self.model, self.encoder_name).config._attn_implementation = "sdpa_without_vmap"
```
kimishpatel: why not use _register_attention_mask?

optimum/exporters/executorch/integrations.py
```diff
@@ -24,15 +24,274 @@
...
+
+            if isinstance(self.model, VoxtralForConditionalGeneration):
+                encoder = VoxtralEncoderExportableModule(self.model)
```
kimishpatel, line 291: I wonder if you just do `encoder_fn = self.model.get_audio_embeds` and then later in `torch.export.export` have `args = (self.model,)`, would that work? I would imagine all audio models have `get_audio_embed` method? 

I think the way we want to generalize this is to rely on specific method names from transformers (and hopefully we can agree on what these would be with hf), e.g. for vision models it is `get_image_features` and `model.vision_tower`? Similarly for audio and then we can generalize this class to export apprpriate components

optimum/executorch/attentions/custom_sdpa.py
```diff
@@ -56,13 +52,16 @@ def custom_sdpa_with_start_pos_forward(
...
     else:
         attn_mask = None
-        # Calculate the input pos from attention mask.
-        # Branch out for float vs bool mask
-        # assert attention_mask.dim() == 2, f"attention_mask must be a 2D matrix."
-        attention_mask = attention_mask.reshape(-1, max_seq_len)
-        first_row_mask = attention_mask[0, :]
-        # [0, 0, 0, 0, -inf, -inf, -inf, -inf], start_pos = 3
-        start_pos = torch.argmin(first_row_mask.to(torch.long)).item() - 1
+        if is_causal:
```
kimishpatel, line 55: this is fine.

optimum/executorch/modeling.py
```diff
@@ -1139,3 +1139,215 @@ def transcribe(
...
+        vocab_size (`int`):
+            Size of the model vocabulary.
+    """
+
+    # Using general `AutoModel` since it usually routes to the correct model variant and there is no
+    # auto model class that captures both audio and image.
+    auto_model_class = AutoModel
+
+    def __init__(self, models: Dict[str, "ExecuTorchModule"], config: "PretrainedConfig"):
+        super().__init__(models=models, config=config)
```
kimishpatel, line 1175: this will register not self.model but self.<decoder> self.<token_embedding> etc, right? https://github.com/huggingface/optimum-executorch/blob/main/optimum/executorch/modeling.py#L98-L103C31. So not sure how this is working

optimum/executorch/modeling.py
kimishpatel: I dont quite get this because encoder, decoder and all are serialized as separate artifacts. so how does this actually work?

optimum/executorch/modeling.py
```diff
@@ -1139,3 +1139,215 @@ def transcribe(
...
+        for required_method in required_methods:
+            if required_method not in self.model.method_names():
+                raise ValueError(
+                    f"Exported .pte file needs to contain the following required methods: {required_methods}"
+                )
+
+        self.encoder_name = None
+        for method_name in self.model.method_names():
+            if method_name == "audio_encoder":
+                self.encoder_name = "audio_encoder"
```
kimishpatel, line 1186: for a truly multimodal you want both vision and uadio

optimum/executorch/modeling.py
kimishpatel: aren't you suppose to combine audio embedding with text for multimodal, depending on the mask?

tests/models/test_modeling_voxtral.py
```diff
@@ -0,0 +1,181 @@
...
+                    cache_position=torch.tensor([pos], dtype=torch.long),
+                )
+            )
+            token = torch.argmax(logits[:, -1, :])
+            print(tokenizer.decode([token.item()]), end="")
+            tokens.append(token.item())
+            pos += 1
+
+        output = tokenizer.decode(tokens, skip_special_tokens=True)
+        self.assertTrue("tattoo" in output)
```
kimishpatel, line 266: shouldnt we check against what hf transformer output is?

tests/models/test_modeling_voxtral.py
```diff
@@ -0,0 +1,181 @@
...
+                .module()
+                .forward(
+                    inputs_embeds=token_embedding,
+                    cache_position=torch.tensor([pos], dtype=torch.long),
+                )
+            )
+            token = torch.argmax(logits[:, -1, :])
+            print(tokenizer.decode([token.item()]), end="")
+            tokens.append(token.item())
+            pos += 1
```
kimishpatel, line 128: we should use the generate method from modeling.py class?

optimum/executorch/modeling.py
jackzhxng: Yeah, just for now we only do one. I would like to do multi modality but I think it's a bit to early to identify the patterns we need to support it in Optimum

optimum/executorch/modeling.py
jackzhxng: I think should be fine since the method will error out with the appropriate error code if there are any shape mismatches already

optimum/executorch/modeling.py
jackzhxng: There will just be one model.pte, with multiple methods

optimum/executorch/modeling.py
jackzhxng: Oh you can see in `xnnpack.py`, we serialize everything to different methods in the same pte artifact

jackzhxng: (This is outdated, check new version)

optimum/exporters/executorch/integrations.py
jackzhxng: The encoder code needs to be in an `nn.module` to export

optimum/exporters/executorch/integrations.py
jackzhxng: Not that I know of, didn't need any of them to do the export

optimum/exporters/executorch/integrations.py
jackzhxng: This is the way they calculate the chunk length in the voxtral code https://github.com/huggingface/transformers/blob/main/src/transformers/models/voxtral/modeling_voxtral.py#L342

optimum/exporters/executorch/integrations.py
jackzhxng: Voxtral pads to a multiple of 30 for encoder batch size

optimum/exporters/executorch/integrations.py
jackzhxng: Addressed above in comment on the `VoxtralEncoderExportableModule `

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
jackzhxng: There's currently no appropriate task to use for it, I put up an issue in Transformers here - https://github.com/issues/created?issue=huggingface%7Ctransformers%7C40302

kimishpatel: ok so you are exporting this manually i guess without using task based stuff

tests/models/test_modeling_voxtral.py
jackzhxng: This is just for ExportedProgram

tests/models/test_modeling_voxtral.py
jackzhxng: Yeah, I think this is a lot less brittle though. I didn't include `check_multimodal_output_quality` also because it's already tested on the pte and this test is more for just development purposes

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
jackzhxng: @zucchini-nlp maybe?

optimum/exporters/executorch/integrations.py
jackzhxng: Added comment

optimum/exporters/executorch/integrations.py
jackzhxng: Hmm what shape are you saying you expect? Maybe for Voxtral it's just different, but I'm pretty confident these are the correct shapes

optimum/exporters/executorch/integrations.py
jackzhxng: Yeah, just need to have that in a `forward`. Also don't know if we can rely on the function name `get_audio_embeds` across different audio models.

optimum/exporters/executorch/integrations.py
jackzhxng: There have been a bunch of comments saying this exact same thing, let's continue the discussion in those above comment threads?

optimum/exporters/executorch/integrations.py
jackzhxng: Yeah because it needs to be able to take in image embeddings too, so the token -> text embeddings part happens separately

optimum/exporters/executorch/integrations.py
jackzhxng: Explained above

optimum/exporters/executorch/integrations.py
jackzhxng: Even if that works, I don't think there's any hard guarantee on what the method name is going to be called, different models already vary in what their audio/vision models attributes are named, such as `vision_tower` and `vision_model`

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
jackzhxng: It works with a fake task name like "multimodal-text-to-text" actually

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
jackzhxng: This is from `causal_lm.py` task, don't have much context over what it's about but just copying over

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
jackzhxng: Copied over from `causal_lm.py`

optimum/exporters/executorch/integrations.py
jackzhxng: Definitely, will follow up on this one in another pr don't want to crowd this one too much

optimum/executorch/modeling.py
kimishpatel: yeah but where does that happen. I couldnt find that in optimum

kimishpatel: ok i see the changed code in xnnpack recipe

optimum/exporters/executorch/integrations.py
kimishpatel: I would have expected [B, features, seq_len] for conv1d. Seqlen is what should encode audio len.

Also divide by 30 is also strange. not clear if it should come from config or not.

optimum/exporters/executorch/integrations.py
kimishpatel: I think we already are relying on get_token_embedding and other things of that nature, so there is already a precedence. Plus we are also relying on model structure elsewhere so I would say confirm with hf folks and use that method instead of copy pasting

optimum/exporters/executorch/integrations.py
kimishpatel: well we are already relying on audio_tower and vision_tower and going as far as relying on conv1 and conv2 submodules (which I definitely dont think we should). So I dont think this is going too far. In fact it allows us to not have to copy paste code for every multimodal we enable, othwerise whats the point of supporting hf model definitions out of the box.

If the issue is lack of stability on method names/attributes, I think we should consult hf folks and settle that. Otherwise this wont scale

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
```diff
@@ -0,0 +1,179 @@
...
+    Args:
+        model: The loaded model instance
+
+    Returns:
+        tuple: (decoder_name, audio_encoder_name, vision_encoder_name)
+    """
+    POTENTIAL_DECODER_NAMES = [
+        "language_model",
+        "text_model",
+    ]
```
zucchini-nlp, line 40: for this one `model.get_decoder()` will work :)

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
zucchini-nlp: not sure what we are trying to achieve here? The generation params of a Multimodal model are the ones used in `model.generation_config` usually, not the ones from `language_model.generation_config` (exception BLIP family which can't be refactored due to BC)

And therefore `from_pretrained` overrides the `model.generation_config`

jackzhxng: So `from_pretrained` only overrides the generation_config.json in `model.generation_config`, but `model.language_model.generation_config` still contains the contents of generation_config.json. I would like for `from_pretrained` to be able to override both.

optimum/exporters/executorch/integrations.py
jackzhxng: Voxtral represents total length with B * seq_len in padded up to multiples of seq_len, you can read more about it in the comment above on line 64

optimum/exporters/executorch/integrations.py
jackzhxng: I don't think we will ever get to a point where we don't have to do a little bit of manual work for each model enabled, although I do wish that were the case. So I think it's reasonable to have model specific code in Optimum as long as it's kept to a minimum.

With regards to this `get_audio_embeds` function specifically, I don't think there's any standardization around this. For example Qwen2.5 audio just uses the audio encoder directly here - https://github.com/huggingface/transformers/blob/7d88f57fc6892b9b3d0092c53e27ae033f1bebc8/src/transformers/models/qwen2_audio/modeling_qwen2_audio.py#L802

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
zucchini-nlp: Hmm, we haven't been updating it because LM's config is basically not used anywhere.It makes sense to have tied generation configs in case of MLLMs, but we can't guarantee that the two configs won't diverge if manually re-set without breaking BC

Btw, can we not use `model.generation_config` when exporting MLLMs, so that t is aligned with general usage? I

jackzhxng: Makes sense that you are concerned about BC, no worries then. We just need to do this here because we pass the text model portion (not the entire model) into `TorchExportableModuleForDecoderOnlyLM`, which uses the generation config to get things such as max cache len.

optimum/exporters/executorch/integrations.py
kimishpatel: > we will ever get to a point where we don't have to do a little bit of manual work for each model enabled,

I think the whole premise of day-0 model enablement was to not have this become the bottleneck. If you realistically feel that we cannot do this, thats fair, but then we should definitely consider what we are getting through this approach rather than copy pasting the model from hf and retrofitting it as we need. Then just use hf checkpoint and convert it as needed.

There are some good learnings, it seems, through the effort of enabling audio and vision model and I think ew should consolidate the learning to make sure the approach of optimum-et still makes sense

optimum/exporters/executorch/integrations.py
kimishpatel: One of my question is not about it being multiple of some values, 30 seconds in this case. I am asking whether batch dim (0th dim) is the right one for sequence dynamism. I am surprised because usually batch dim is not where seq length is specified. Is there a test code pointer for voxtral in hf transformers that i can refer to? Problem with this might be that, if it is wrong, you might find out about it in a bit of hard-to-debug way

jackzhxng: Hmm I see. Yes, there's a comment to a code pointer in HF above right above this explaining this batch dim stacking mechanism

optimum/exporters/executorch/integrations.py
jackzhxng: This was necessary due to some torch.export nuances, it won't specialize on special token at position 1 if you do this. Regardless, this was removed since we now do split prefill

optimum/exporters/executorch/integrations.py
kimishpatel: Aah. interesting. ok that makes sense.

kimishpatel: This is specific to voxtral I imagine. So in some sense export shape also needs to be known from model itself which also means that the model API is not the same. Meaning another audio LM may handle this differently?

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
kimishpatel: @jackzhxng do you wanna fix this?

tests/models/test_modeling_voxtral.py
kimishpatel: I understand but can you not do generate on the wrapped module that wraps exportedPrograms?

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
jackzhxng: @zucchini-nlp I can explore this more as a follow up, but for now this doesn't work out of the box since `.get_decoder()` returns `LlamaModel`, whereas I need the `LlamaModelForCausalLM` which includes extra components necessary for the text decoder such as the `lm_head`

zucchini-nlp: Ah I see, voxtral is kinda different from other model and it has the head glued to LM. Yeah, let's see how it works with vision models, because the standard format is to save an lm head separate from backbone encoders

PR #128, opened 2025-08-25
Bump ET to 0.7 in pip package

jackzhxng:


jackzhxng:
Merging - other tests are pre-existing and gemma tests seem to be flakes

PR #136, opened 2025-09-04
Bump transformers to 4.56.1

jackzhxng:
Some changes needed for the bump:
- Use early initialization to populate cache layer info - https://github.com/huggingface/transformers/pull/39797
- No more max seq length and batch size args in DecoderOnlyLM constructor
- Hybrid cache is getting deprecated, with static cache serving the same purpose with per-layer configs - https://github.com/huggingface/transformers/pull/40276

### DIFF
--- a/optimum/executorch/attentions/custom_kv_cache.py
+++ b/optimum/executorch/attentions/custom_kv_cache.py
@@ -37,7 +37,6 @@ class ETCustomStaticCache(StaticCache):
         max_cache_len: Optional[int] = None,
         device: Union[torch.device, str, None] = None,
         dtype: torch.dtype = torch.float32,
-        layer_device_map: Optional[Dict[int, Union[str, torch.device, int]]] = None,
     ):
         super().__init__(
             config=config,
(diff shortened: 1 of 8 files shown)


kimishpatel (approved):

optimum/executorch/attentions/custom_kv_cache.py
```diff
@@ -37,22 +37,23 @@ def __init__(
...
             config=config,
             max_batch_size=max_batch_size,
             max_cache_len=max_cache_len,
             device=device,
             dtype=dtype,
-            layer_device_map=layer_device_map,
+        )
+        num_heads = getattr(config, "num_key_value_heads", config.num_attention_heads)
+        head_dim = getattr(config, "head_dim", config.hidden_size // config.num_attention_heads)
+        self.early_initialization(
```
kimishpatel, line 50: what is this doing?

jackzhxng: (In pr description)

optimum/exporters/executorch/integrations.py
```diff
@@ -475,8 +480,11 @@ def export(
             mutated_gm = RemoveRedundantTransposes()(exported_program.module())[0]
             exported_program = torch.export.export(
                 mutated_gm,
-                args=(input_ids, cache_position),
-                kwargs={},
+                args=(),
+                kwargs={
+                    "input_ids": input_ids,
+                    "cache_position": cache_position,
+                },
```
kimishpatel, line 487: why this change

optimum/executorch/attentions/custom_kv_cache.py
kimishpatel: linked pr description says that cache_position will also be removed. Is it talking about cache maintaining cache_position? Something to keep track of

optimum/exporters/executorch/integrations.py
jackzhxng: https://github.com/huggingface/transformers/pull/39836/files#diff-6a4faa71e9436fb9187908174ea7ae18774e4ad3ed0cda17a29edec657383a41L249-R254

This changed from args to kwargs in transformers

optimum/executorch/attentions/custom_kv_cache.py
jackzhxng: Oh it's the part on early_initialization - basically you need to call this to initialize some of the attributes you need on each cache layer

PR #138, opened 2025-09-10
Add quantization options for encoder and group size

jackzhxng:
### Summary
- Quantization arg for encoder
- Group size args for decoder, encoder, and embedding

Stack (oldest at bottom):
* #139 
* #138

### DIFF
--- a/optimum/exporters/executorch/quantization.py
+++ b/optimum/exporters/executorch/quantization.py
@@ -36,11 +36,17 @@ def quantize_model_(
     )
     from torchao.utils import unwrap_tensor_subclass
 
-    if qembedding_group_size == 0:
-        embedding_weight_granularity = PerAxis(0)
-    else:
-        embedding_weight_granularity = PerGroup(qembedding_group_size)
     if qembedding_config:
+        if qlinear_config == "8w":
+            assert (
+                qembedding_group_size == 0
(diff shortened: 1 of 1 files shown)


optimum/exporters/executorch/quantization.py
```diff
@@ -32,16 +36,20 @@ def quantize_model_(
     )
     from torchao.utils import unwrap_tensor_subclass
 
+    if qembedding_group_size == 0:
+        embedding_weight_granularity = PerAxis(0)
+    else:
+        embedding_weight_granularity = PerGroup(qembedding_group_size)
```
mergennachin: assert qembedding_group_size is a postive number power of 2?

optimum/exporters/executorch/quantization.py
```diff
@@ -32,16 +36,20 @@ def quantize_model_(
     )
     from torchao.utils import unwrap_tensor_subclass
 
+    if qembedding_group_size == 0:
+        embedding_weight_granularity = PerAxis(0)
```
mergennachin: assert that qembedding_config is either None or "8w" in the if case

optimum/exporters/executorch/quantization.py
```diff
@@ -32,16 +36,20 @@ def quantize_model_(
...
+    else:
```
mergennachin: assert that qembedding_config must be 4w in the else case

PR #142, opened 2025-09-11
Enable exporting Voxtral on MPS

manuelcandales:


optimum/executorch/attentions/custom_kv_cache.py
```diff
@@ -50,7 +50,7 @@ def __init__(
 
         # make sure layer_device_map is none
         assert layer_device_map is None
-        assert device is None or device == "cpu", "Device must be None or 'cpu'"
+        # assert device is None or device == "cpu", "Device must be None or 'cpu'"
```
larryliu0820: Can you assert it's one of `cpu` or `mps`

larryliu0820 (approved):

PR #144, opened 2025-09-18
Add Granite Speech 3.3

jackzhxng:
```
Certainly! Here's the transcribed written format of Timothy's actions and thoughts:

After his nap, Timothy leisurely stretched, first one gray velvet foot, then the other. He then slowly rolled, indolently, to his plate. Turning
⚠️ DISCLAIMER: Python-based perf measurementstare approximaie and may not match absolute speeds on Android/iOS apps. They are intended for relative comparisons—-e.g. SDPA vs. custom SDPA, FP16 vs. FP32—-so you can
gauge performance improvements from each optimization step. For end-to-end, platform-accurate benchmarks, please use the official ExecuTorch apps:

PyTorchObserver {"prompt_tokens": 239, "generated_tokens": 64, "model_load_start_ms": 0, "model_load_end_ms": 0, "inference_start_ms": 1758213608046, "token_encode_end_ms": 1758213608207, "model_execution_start_ms"
: 0, "model_execution_end_ms": 0, "inference_end_ms": 1758213621530, "prompt_eval_end_ms": 1758213611648, "first_token_ms": 1758213611824, "aggregate_sampling_time_ms": 13307, "SCALING_FACTOR_UNITS_PER_SECOND": 100
0}
        Prompt Tokens: 239 Generated Tokens: 64
        Total inference time:           13.484000 (seconds)              Rate:  4.746366 (tokens/second)
                Prompt evaluation:      3.602000 (seconds)               Rate:  66.352027 (tokens/second)
                Generated 64 tokens:    9.882000 (seconds)               Rate:  6.476422 (tokens/second)
        Time to first generated token:  3.778000 (seconds)
        Sampling time over 303 tokens:  13.307000 (seconds)
```

### DIFF
--- a/optimum/exporters/executorch/integrations.py
+++ b/optimum/exporters/executorch/integrations.py
@@ -147,7 +147,11 @@ class AudioExportableModule(torch.nn.Module):
         self,
         input_features: torch.FloatTensor,
     ):
-        audio_embeds = self.model.get_audio_features(input_features)
+        # TODO: remove on next Transformers pin bump.
+        if hasattr(self.model, "get_audio_embeds"):
+            audio_embeds = self.model.get_audio_embeds(input_features)
+        else:
+            audio_embeds = self.model.get_audio_features(input_features)
         return audio_embeds.unsqueeze(0)
 
 


optimum/exporters/executorch/tasks/multimodal_text_to_text.py
```diff
@@ -39,6 +39,7 @@ def _validate_multimodal_components(model):
         "text_model",
     ]
     POTENTIAL_AUDIO_ENCODER_NAMES = [
+        "encoder",  # Here mainly for Granite Speech.
```
mergennachin, line 42: cc @zucchini-nlp fyi, there are some one-off things like this

optimum/exporters/executorch/integrations.py
```diff
@@ -88,30 +88,46 @@ def prepare_export_inputs(self):
...
+                }
+            ]
+            processed_inputs = apply_chat_template_with_fallback(
+                processor,
+                sample_conversation_with_audio,
+                add_generation_prompt=True,
+                tokenize=True,
+                return_dict=True,
+                return_tensors="pt",
+            )
```
mergennachin: can this be replaced by calling process_conversation_inputs?

jackzhxng: Thought about it, think I'd like to keep the export separate, also if I used process_conversation_inputs I would still need to have this conditional here since I want to use the sample input straight from Granite's repo

optimum/exporters/executorch/utils.py
```diff
@@ -126,3 +130,67 @@ def verify_eos_tokens_in_pretrained_tokenizer(model_eos_ids: List[int], tokenize
...
+        if audio_path is None:
+            raise ValueError("No audio content found in conversation for GraniteSpeechProcessor")
+
+        # Download and process audio
+        resp = requests.get(audio_path)
+        resp.raise_for_status()
+        buf = io.BytesIO(resp.content)
+
+        wav, sampling_rate = torchaudio.load(buf, normalize=True)
+        assert wav.shape[0] == 1 and sampling_rate == 16000  # mono, 16khz
```
mergennachin: What happens if the wav file is not adhering to this contract? 

```
  if wav.shape[0] != 1:
      wav = wav.mean(dim=0, keepdim=True)  # Convert stereo to mono
      logger.warning(f"Resampled audio stereo to mono")
  if sampling_rate != 16000:
      wav = torchaudio.functional.resample(wav, sampling_rate, 16000)
      logger.warning(f"Resampled audio from {sampling_rate}Hz to 16000Hz")
  ```

File: optimum/exporters/executorch/utils.py [hunk 4]
```diff
@@ -126,3 +130,67 @@ def verify_eos_tokens_in_pretrained_tokenizer(model_eos_ids: List[int], tokenize
...
+        conversation = copy.deepcopy(input_conversation)
+        audio_path = None
+
+        # Extract audio content and remove from conversation
+        for i, item in enumerate(conversation):
+            if "type" in item and item["type"] == "audio":
+                audio_path = item["content"]
+                # Remove the audio content from the input conversation since it
+                # is handled outside for Granite.
+                conversation.pop(i)
```
mergennachin: You don't want to remove modify during iteration.


```
  audio_items = [(i, item) for i, item in enumerate(conversation)
                 if item.get("type") == "audio"]
  if audio_items:
      idx, audio_item = audio_items[0]
      audio_path = audio_item["content"]
      del conversation[idx]
  else:
      raise ValueError("No audio content found in conversation")
```

optimum/exporters/executorch/utils.py
```diff
@@ -126,3 +130,67 @@ def verify_eos_tokens_in_pretrained_tokenizer(model_eos_ids: List[int], tokenize
...
+                break
+
+        if audio_path is None:
+            raise ValueError("No audio content found in conversation for GraniteSpeechProcessor")
+
+        # Download and process audio
+        resp = requests.get(audio_path)
+        resp.raise_for_status()
+        buf = io.BytesIO(resp.content)
```
mergennachin: Protect with try catch

optimum/exporters/executorch/utils.py
```diff
@@ -126,3 +130,67 @@ def verify_eos_tokens_in_pretrained_tokenizer(model_eos_ids: List[int], tokenize
...
+    prior to feeding into the processor.
+
+    Args:
+        processor: The processor to use for input processing
+        tokenizer: The tokenizer to use for text processing
+        input_conversation: List of conversation messages, may contain audio content
+
+    Returns:
+        Processed inputs ready for model consumption
+    """
```
mergennachin, line 155: validate the input_conversation is in fact non-empty List[Dict]

optimum/exporters/executorch/utils.py
mergennachin: Can't we have an option to just use a local audio_file? Always connecting to the internet is not great

optimum/exporters/executorch/integrations.py
```diff
@@ -88,30 +88,46 @@ def prepare_export_inputs(self):
...
-        processed_inputs = apply_chat_template_with_fallback(
-            processor,
-            sample_conversation_with_audio,
-            add_generation_prompt=True,
-            tokenize=True,
-            return_dict=True,
-            return_tensors="pt",
-        )
+
+        if model_id.startswith("ibm-granite/granite-speech"):
```
mergennachin: there could be other fine-tuned granite model besides ibm domain..

What about something like this?
```
  if hasattr(self.model.config, 'encoder_config') and \
     self.model.config.encoder_config.model_type == 'granite_speech_encoder':
 ```

File: optimum/exporters/executorch/utils.py [hunk 8]
```diff
@@ -126,3 +130,67 @@ def verify_eos_tokens_in_pretrained_tokenizer(model_eos_ids: List[int], tokenize
     is_valid = any(model_id in candidate_eos_ids for model_id in model_eos_ids)
 
     return is_valid
+
+
+def process_conversation_inputs(
+    processor: ProcessorMixin,
+    tokenizer: PreTrainedTokenizer,
+    input_conversation: List[Dict],
```
mergennachin: `List[Dict[str, Any]],`

optimum/exporters/executorch/utils.py
mergennachin:
```
if audio_path.startswith(('http://', 'https://'))
  try 
     requests.get(audio_path)
   catch
   ....
   wav, sampling_rate = torchaudio.load(audio_path, normalize=True)
elif os.path.exists(audio_path):
   wav, sampling_rate = torchaudio.load(audio_path, normalize=True)
else:

Comment (jackzhxng) [reply optimum/exporters/executorch/utils.py]: I think we can do this in another pr, also I'd prefer to stay consistent with export inputs

File: optimum/exporters/executorch/utils.py (hunk 6 above)
Comment (jackzhxng) [reply optimum/exporters/executorch/utils.py:155]: Empty should be ok?

File: optimum/exporters/executorch/integrations.py [hunk 9]
```diff
@@ -88,30 +88,46 @@ def prepare_export_inputs(self):
...
+            import torchaudio
+            from huggingface_hub import hf_hub_download
+
+            audio_path = hf_hub_download(repo_id=model_id, filename="10226_10111_000000.wav")
+            wav, _sampling_rate = torchaudio.load(audio_path, normalize=True)
+            processed_inputs = processor(
+                "",  # No text needed.
+                wav,
+                return_tensors="pt",
+            )
```
zucchini-nlp, line 104: is this workaround caused by HF implementation or the model has no chat template?

optimum/exporters/executorch/tasks/multimodal_text_to_text.py
zucchini-nlp: with audio models there is no standard yet tbh. We'll go over audio modality in the future releases

optimum/exporters/executorch/integrations.py
```diff
@@ -88,30 +88,46 @@ def prepare_export_inputs(self):
...
-        processed_inputs = apply_chat_template_with_fallback(
-            processor,
-            sample_conversation_with_audio,
-            add_generation_prompt=True,
-            tokenize=True,
-            return_dict=True,
-            return_tensors="pt",
-        )
+
+        if model_id.startswith("ibm-granite/granite-speech-3.3"):
```
mergennachin: is there a way to not special case specifically on "ibm-granite/granite-speech-3.3"? 

in my previous comment, what i meant was that if someone finetunes a granite model it will be somethign like jack/granite-speech-3.3, and it won't work.

are there other ways you can check whether it is granite speech model somehow?

optimum/exporters/executorch/integrations.py
jackzhxng: Maybe - processor's apply_chat_template didn't work, so I looked at the example here on the official repo and followed how they did it: https://huggingface.co/ibm-granite/granite-speech-3.3-2b

optimum/exporters/executorch/integrations.py
jackzhxng: Oh, i see. yeah can do

mergennachin (approved):

PR #146, opened 2025-09-18
Fix whisper caching issue

JacobSzwejbka:
This PR is also reliant on some pin bumps happening in @jackzhxng's prs. The formatting changes happened automatically when I ran make style.

Whisper was using a deprecated api pattern from transformers. This PR switches it to a style used in the Qualcomm scripts that seems to work well.

### DIFF
--- a/optimum/exporters/executorch/integrations.py
+++ b/optimum/exporters/executorch/integrations.py
@@ -543,7 +543,7 @@ class VisionEncoderExportableModule(torch.nn.Module):
                     self.model,
                     args=(),
                     kwargs={"pixel_values": pixel_values},
-                    strict=True,
+                    strict=False,
                 )
             }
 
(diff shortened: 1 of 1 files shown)


jackzhxng (commented):
Thanks @JacobSzwejbka. This should be able to fix T5 as well.

optimum/executorch/modeling.py
```diff
@@ -495,7 +524,12 @@ def forward(
             encoder_outputs = self.encoder.forward((input_ids,))[0]
             self.stats.on_prompt_eval_end()
 
-        result = (self.decoder.forward((decoder_input_ids, encoder_outputs, cache_position))[0], encoder_outputs)
+        result = (
+            self.decoder.forward((decoder_input_ids, encoder_outputs, cache_position))[
```
jackzhxng:
```suggestion
            self.text_decoder.forward((decoder_input_ids, encoder_outputs, cache_position))[
```

PR #147, opened 2025-09-19
Modify image embedding return for Llava compatibility

larryliu0820:
For `Gemma3` we don't need to unsqueeze the image embedding. For `Llava` we got a list of 2D tensors and here I'm assuming it only contains 1 tensor.

https://github.com/huggingface/transformers/blob/main/src/transformers/models/llava/modeling_llava.py#L214

### DIFF
--- a/optimum/exporters/executorch/integrations.py
+++ b/optimum/exporters/executorch/integrations.py
@@ -77,7 +77,12 @@ class VisionExportableModule(torch.nn.Module):
         input_features: torch.FloatTensor,
     ):
         image_embeds = self.model.get_image_features(input_features)
-        return image_embeds.unsqueeze(0)
+        # Hack, assuming the text decoder will take a 3D tensor (batch_size, seq_len, hidden_size)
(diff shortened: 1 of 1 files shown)


optimum/exporters/executorch/integrations.py
```diff
@@ -76,7 +76,12 @@ def forward(
         input_features: torch.FloatTensor,
     ):
         image_embeds = self.model.get_image_features(input_features)
-        return image_embeds.unsqueeze(0)
+        # Hack, assuming the text decoder will take a 3D tensor (batch_size, seq_len, hidden_size)
+        if "llava" in self.model.config.name_or_path:
```
jackzhxng: What about something like

```
if isinstance(image_embeds, List):
    image_embeds = torch.stack(image_embeds)
else:
    image_embeds.unsqueeze(0)
```

larryliu0820: Let me try, I don't think this is exportable though

kimishpatel (approved):

PR #158, opened 2025-10-01
Add CUDA recipe

larryliu0820:
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/optimum/exporters/executorch/recipes/cuda.py
+++ b/optimum/exporters/executorch/recipes/cuda.py
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 import logging
-import torch
 from typing import Dict, Union
 
+import torch
 from tabulate import tabulate
 from torch.export import ExportedProgram
 from torch.nn.attention import SDPBackend
@@ -35,6 +35,7 @@ from ..integrations import (
     Seq2SeqLMExportableModule,
 )
 from ..recipe_registry import register_recipe
+
 
 aten = torch.ops.aten
 


optimum/commands/export/executorch.py
```diff
@@ -126,6 +126,20 @@ def parse_args_executorch(parser):
...
+        "--dtype",
+        type=str,
+        choices=["float32", "float16", "bfloat16"],
+        required=False,
+        help="Data type for model weights. Options: float32, float16, bfloat16. Default: float32. For quantization (int8/int4), use the --qlinear arguments.",
+    )
+    required_group.add_argument(
+        "--device",
+        type=str,
+        choices=["cpu", "cuda", "cuda:0", "cuda:1", "cuda:2", "cuda:3"],
```
kimishpatel, line 139: why 0, 1, 2 etc? that seems like a strange thing

optimum/exporters/executorch/recipes/cuda.py
```diff
@@ -0,0 +1,127 @@
...
+from torch.export import ExportedProgram
+from torch.nn.attention import SDPBackend
+
+from executorch.devtools.backend_debug import get_delegation_info
+from executorch.exir import (
+    EdgeCompileConfig,
+    ExecutorchProgram,
+    to_edge_transform_and_lower,
+)
+from optimum.executorch.passes.remove_padding_idx_embedding_pass import RemovePaddingIdxEmbeddingPass
```
kimishpatel, line 29: needed?

optimum/exporters/executorch/recipes/cuda.py
```diff
@@ -0,0 +1,127 @@
...
+        if len(exported_programs) == 1:
+            exported_programs = {"forward": next(iter(exported_programs.values()))}
+
+        # CUDA backend compile spec with method name.
+        partitioners = {key: [CudaPartitioner([CudaBackend.generate_method_name_compile_spec(key)])] for key in exported_programs.keys()}
+        # Add decompositions for triton to generate kernels.
+        for key, ep in exported_programs.items():
+            exported_programs[key] = ep.run_decompositions({
+              aten.conv1d.default: conv1d_to_conv2d,
+            })
```
kimishpatel: why does htis need to happen manually?

optimum/exporters/executorch/recipes/cuda.py
```diff
@@ -0,0 +1,127 @@
...
+        return {pte_name: et_prog}
+
+    # Decomposes SDPA since we don't have a flash attention kernel for it yet.
+    with torch.nn.attention.sdpa_kernel([SDPBackend.MATH]), torch.no_grad():
+        exported_progs = model.export()
+
+    if (
+        model.config._attn_implementation == "custom_sdpa"
+        or model.config._attn_implementation == "custom_sdpa_ring_kv_cache"
+    ):
```
kimishpatel, line 124: how are we running custom sdpa for cuda? or is it supposed to be a graph break?

optimum/exporters/executorch/recipes/cuda.py
```diff
@@ -0,0 +1,127 @@
...
+                f"\nDelegation info for {pte_name}.pte: {tabulate(delegation_info.get_operator_delegation_dataframe(), headers='keys', tablefmt='fancy_grid')}"
+            )
+        return {pte_name: et_prog}
+
+    # Decomposes SDPA since we don't have a flash attention kernel for it yet.
+    with torch.nn.attention.sdpa_kernel([SDPBackend.MATH]), torch.no_grad():
+        exported_progs = model.export()
+
+    if (
+        model.config._attn_implementation == "custom_sdpa"
```
kimishpatel, line 122: this is a generic recipe. does all supported transformer models have _attn_implementation? I dont suppose you are introducing this new behavior but just asking

kimishpatel (changes requested):
any tests?

optimum/commands/export/executorch.py
larryliu0820: For an environment with more than 1 GPU you can specify which GPU

optimum/exporters/executorch/recipes/cuda.py
larryliu0820: triton doesn't generate conv1d kernel. This decomposition route conv1d to conv2d to make sure we have a kernel

optimum/exporters/executorch/recipes/cuda.py
larryliu0820: Used in line 96

optimum/exporters/executorch/recipes/cuda.py
larryliu0820: right this doesn't work. I can remove the whole block. Or actually it should throw an error for now.

optimum/commands/export/executorch.py
kimishpatel: is that a normal thing to do. I am a bit surprised as to where something like this will be used. Especially optimum-executorch is really for generating aot artifact and we dont know where it will be deployed

larryliu0820: > Especially optimum-executorch is really for generating aot artifact and we dont know where it will be deployed

I think this should be fine, I don't think the aoti compiled artifact has this GPU index hardcoded in. I think this is convenient if you have 2 GPUs and the first one is occupied / doesn't have enough memory and we want to export on the second GPU.

kimishpatel: Sure. I do find it a bit awkward but sure.

larryliu0820:
> would be good to have a test

Added a test. The thing for test is that it has to run on CI with cuda so it's probably being skipped. I think it would be good to add CI in ET.

tests/models/test_modeling_voxtral.py
```diff
@@ -324,3 +326,28 @@ def test_voxtral_audio_text_to_text_generation_with_custom_sdpa_kv_cache_8da4w_8
...
+                f"--model {model_id} "
+                f"--task {task} "
+                f"--recipe {recipe} "
+                "--dtype bfloat16 "
+                "--device cuda:0 "
+                "--max_seq_len 1024 "
+                f"--output_dir {output_dir}"
+            )
+            subprocess.run(cmd, shell=True, check=True)
+            self.assertTrue(os.path.exists(os.path.join(output_dir, "model.pte")))
```
mergennachin, line 353: can you also test the existence of .ptd file too

optimum/exporters/executorch/convert.py
```diff
@@ -86,5 +86,6 @@ def export_to_executorch(
             logging.info(
                 f"Saved exported program to {full_path} ({os.path.getsize(full_path) / (1024 * 1024):.2f} MB)"
             )
+        prog.write_tensor_data_to_file(output_dir)
```
mergennachin, line 89: are you sure that we want this behavior for all existing recipes?

optimum/exporters/executorch/recipes/cuda.py
```diff
@@ -0,0 +1,129 @@
...
+
+@register_recipe("cuda")
+def export_to_executorch_with_cuda(
+    model: Union[
+        CausalLMExportableModule,
+        MaskedLMExportableModule,
+        Seq2SeqLMExportableModule,
+        MultiModalTextToTextExportableModule,
+    ],
+    **kwargs,
```
mergennachin, line 51: dtype and device from the cli is never used

optimum/exporters/executorch/recipes/cuda.py
```diff
@@ -0,0 +1,129 @@
...
+                f"\nDelegation info for {pte_name}.pte: {tabulate(delegation_info.get_operator_delegation_dataframe(), headers='keys', tablefmt='fancy_grid')}"
+            )
+        return {pte_name: et_prog}
+
+    # Decomposes SDPA since we don't have a flash attention kernel for it yet.
+    with torch.nn.attention.sdpa_kernel([SDPBackend.MATH]), torch.no_grad():
+        exported_progs = model.export()
+
+    if (
+        model.config._attn_implementation == "custom_sdpa"
```
mergennachin, line 122: can you error out early on, before exporting, so that users don't have to wait the whole time

optimum/exporters/executorch/recipes/cuda.py
```diff
@@ -0,0 +1,129 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
```
mergennachin, line 1: You are the author, so add Meta copyright.

optimum/exporters/executorch/convert.py
larryliu0820: If we don't put weight as external then this is just no-op

optimum/exporters/executorch/recipes/cuda.py
larryliu0820: It goes into `MultiModalTextToTextExportableModule` through https://github.com/huggingface/optimum-executorch/blob/main/optimum/commands/export/executorch.py#L176-L179

optimum/exporters/executorch/recipes/cuda.py
larryliu0820: Good point, I can do that

optimum/exporters/executorch/recipes/cuda.py
larryliu0820: I think Hugging Face requires us to add this license.

optimum/exporters/executorch/convert.py
lucylq: Yeah that's right^

optimum/exporters/executorch/recipes/cuda.py
mergennachin: I think you're mixing two things: 

- License -- in this case Apache license, which I agree HF requires us. 
- Copyright header -- they don't have such requirement. See an example here: https://github.com/huggingface/optimum-executorch/blob/dfa2d7ab0bce75d865e923717072b645ed0432b1/optimum/executorch/attentions/custom_kv_cache.py#L1

PR #159, opened 2025-10-03
Address comments on #158

larryliu0820:
* Moved attention check earlier
* Check for .ptd file as well in unit test


mergennachin (approved):
See my reply here https://github.com/huggingface/optimum-executorch/pull/158/files#r2403419621

PR #163, opened 2025-10-10
Add SmolVLM

### LINKED ISSUE #161: SmolVLM
Cannot export due to data dependent loop on image input batch size - https://github.com/huggingface/transformers/blob/main/src/transformers/models/idefics3/modeling_idefics3.py#L148

jackzhxng:
```
 The image depicts a clear and vivid view of a cityscape with a prominent focus on the Statue of Liberty in the foreground. The Statue of Liberty is centrally positioned and prominently displayed, standing on a small, rocky platform that extends into the water. The water in the background is a deep blue, with a few
 ships visible
⚠️ DISCLAIMER: Pythoi-based perf measurements are approximate and may not match absolute speeds on Android/iOS apps. They are intended for relative comparisons——e.g. SDPA vs. custom SDPA, FP16 vs. FP32——so you can gauge performance improvements from each optimization step. For end-to-end, platform-accurate benchmar
ks, please use the official ExecuTorch apps:

PyTorchObserver {"prompt_tokens": 1197, "generated_tokens": 64, "model_load_start_ms": 0, "model_load_end_ms": 0, "inference_start_ms": 1760545277225, "token_encode_end_ms": 1760545277506, "model_execution_start_ms": 0, "model_execution_end_ms": 0, "inference_end_ms": 1760545314593, "prompt_eval_end_ms": 1760545308
404, "first_token_ms": 1760545308515, "aggregate_sampling_time_ms": 37068, "SCALING_FACTOR_UNITS_PER_SECOND": 1000}
        Prompt Tokens: 1197 Generated Tokens: 64
        Total inference time:           37.368000 (seconds)              Rate:  1.712695 (tokens/second)
                Prompt evaluation:      31.179000 (seconds)              Rate:  38.391225 (tokens/second)
                Generated 64 tokens:    6.189000 (seconds)               Rate:  10.340927 (tokens/second)
        Time to first generated token:  31.290000 (seconds)
        Sampling time over 1261 tokens: 37.068000 (seconds)

INFO     root:utils.py:87 Starting perplexity check with multimodal model 'HuggingFaceTB/SmolVLM-Instruct' ...
Fetching 2 files: 100%|███████████████████████████████████████████████████| 2/2 [00:00<00:00, 5349.88it/s]
INFO     root:utils.py:135 ✓ Perplexity check passed: 2.84 <= 5
PASSED
```

Requires the following Transformers changes (also WIP) - https://github.com/huggingface/transformers/pull/41629

### DIFF
--- a/optimum/exporters/executorch/integrations.py
+++ b/optimum/exporters/executorch/integrations.py
@@ -38,59 +38,10 @@ from optimum.executorch.attentions.custom_sdpa import get_custom_sdpa_for_ring_k
 from .utils import apply_chat_template_with_fallback, process_conversation_inputs, save_config_to_constant_methods
 
 
-def _patch_idefics3_vision_embeddings_for_export(vision_model):
-    """
-    Patch Idefics3VisionEmbeddings to make it:
-    - Export-friendly by removing data-dependent operations (forces assumption of image input
-      batch_size = 1, and thus a full attention mask).
-    - Not use aten.bucketize, which has no available decompositions or kernels in ExecuTorch.
-    """
-    import types
-
-    def export_friendly_forward(
-        self, pixel_values: torch.FloatTensor, patch_attention_mask: torch.BoolTensor
-    ) -> torch.Tensor:
-        batch_size, _, max_im_h, max_im_w = pixel_values.shape
-
-        patch_embeds = self.patch_embedding(pixel_values)
-        embeddings = patch_embeds.flatten(2).transpose(1, 2)
-
-        nb_patches_h = max_im_h // self.patch_size
-        nb_patches_w = max_im_w // self.patch_size
-        N = self.num_patches_per_side
-
-        # For export, we assume full attention mask and compute position IDs statically.
-        # This avoids the data-dependent loop over batch dimension.
-        h_indices = torch.arange(nb_patches_h, device=pixel_values.device, dtype=torch.long)
-        w_indices = torch.arange(nb_patches_w, device=pixel_values.device, dtype=torch.long)
-
-        # This replaces bucketize(x, boundaries=[1/N, 2/N, ...], right=True) ≈ floor(x * N), which
-        # we don't have a kernel for at the moment.
-        bucket_coords_h = (h_indices * N) // nb_patches_h
-        bucket_coords_w = (w_indices * N) // nb_patches_w
-
-        bucket_coords_h = torch.clamp(bucket_coords_h, max=N - 1)
-        bucket_coords_w = torch.clamp(bucket_coords_w, max=N - 1)
-
-        pos_ids = (bucket_coords_h[:, None] * N + bucket_coords_w[None, :]).reshape(-1)
-        position_ids = pos_ids.unsqueeze(0).expand(batch_size, -1)
-        embeddings = embeddings + self.position_embedding(position_ids)
-        return embeddings
-
-    # Patch the forward method.
-    vision_model.embeddings.forward = types.MethodType(export_friendly_forward, vision_model.embeddings)
-
-
 class VisionExportableModule(torch.nn.Module):
     def __init__(self, model: torch.nn.Module):
         super().__init__()
         self.model = model
-
-        # Patch Idefics3 vision embeddings to make it exportable.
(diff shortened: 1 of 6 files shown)


optimum/exporters/executorch/integrations.py
```diff
@@ -34,18 +35,63 @@
...
+from .utils import apply_chat_template_with_fallback, process_conversation_inputs, save_config_to_constant_methods
+
+def _patch_idefics3_vision_embeddings_for_export(vision_model):
+    """
+    Patch Idefics3VisionEmbeddings to make it export-friendly by removing data-dependent operations.
+    This assumes batch_size=1 and a full attention mask (all 1s).
+    """
+    import types
+
+    def export_friendly_forward(self, pixel_values: torch.FloatTensor, patch_attention_mask: torch.BoolTensor) -> torch.Tensor:
```
jackzhxng: @zucchini-nlp This part could not export because of the data dependent loop, I unroll it here to just handle one image. Here's the original code -
https://github.com/huggingface/transformers/blob/main/src/transformers/models/idefics3/modeling_idefics3.py#L149

optimum/exporters/executorch/integrations.py
```diff
@@ -34,18 +35,63 @@
...
+        nb_patches_h = max_im_h // self.patch_size
+        nb_patches_w = max_im_w // self.patch_size
+        N = self.num_patches_per_side
+
+        # For export, we assume full attention mask and compute position IDs statically.
+        # This avoids the data-dependent loop over batch dimension.
+        h_indices = torch.arange(nb_patches_h, device=pixel_values.device, dtype=torch.long)
+        w_indices = torch.arange(nb_patches_w, device=pixel_values.device, dtype=torch.long)
+
+        # This replaces bucketize(x, boundaries=[1/N, 2/N, ...], right=True) ≈ floor(x * N), which
```
jackzhxng: This is the second change, we don't have a kernel for `aten.bucketize` so I compute it manually

optimum/exporters/executorch/integrations.py
zucchini-nlp: hmm, i thought it was fixed in https://github.com/huggingface/transformers/pull/39614

jackzhxng: Ah that one just exports the vision encoder, this one exports the [get_image_features](https://github.com/huggingface/transformers/blob/main/src/transformers/models/smolvlm/modeling_smolvlm.py#L587) function which calls the vision encoder. I'm thinking that some code in this function might be confusing to export cc @tugsbayasgalan

zucchini-nlp: ahh i see, prob it is the part where images are unpadded and that code is value dependant. 

For my understanding, can we export the vision and the LM part separately but not the merging logic?  Most `get_image_features` might not be 100% exportable for VLMs, so keeping it outside of `export` can work better on long term

jackzhxng: The reason we export `get_image_features` is that if we only exported the vision encoder, we would need to write the rest of the merging logic in C++ which is difficult to do and harder to scale. I wonder if it's possible to upstream an exportable version of `get_image_features`? If not, I'm happy to just monkey patch like this

zucchini-nlp: > I wonder if it's possible to upstream an exportable version of `get_image_features`

This would be the best solution, and if you want to submit a PR it'll very welcome. I see that we're passing `pixel_attention_mask=None` in all cases in this PR but the vision backbone is still not exportable?

bhu619:
Hi, I saw your #163 PR, which aims to add support for the SmolVLM model to optimum-executorch. I'm currently also trying to export this model under ExecuTorch for on-device deployment, but noticed you've closed this PR. Could you share why it was closed? Was there a technical blocker, or has the support been implemented another way?

Also, if possible, could I still refer to the commits you made before closing, to try exporting the SmolVLM model myself? Any tips or known issues would be really helpful. Thanks!

jackzhxng:
Hey, I'm working on other stuff now so I closed it but feel free to take over here!

PR #164, opened 2025-10-10
Add qlinear_packing_format options

desertfire:
Summary: Using --qlinear_packing_format to apply the following torchao API,
```
Int4WeightOnlyConfig(
    group_size=32,
    int4_packing_format="tile_packed_to_4d",
    int4_choose_qparams_algorithm="hqq",
)
```
This enables aten._weight_int4pack_mm for matrix multiplication.

### DIFF
--- a/optimum/commands/export/executorch.py
+++ b/optimum/commands/export/executorch.py
@@ -95,7 +95,7 @@ def parse_args_executorch(parser):
         choices=["tile_packed_to_4d"],
         required=False,
         help=(
-            "Int4 packing format for encoder linear layers.\n"
+            "Packing format for decoder linear layers.\n"
             "Only applicable to certain backends such as CUDA and Metal\n\n"
             "Options:\n"
             "  tile_packed_to_4d  - int4 4d packing format"
@@ -123,7 +123,7 @@ def parse_args_executorch(parser):
         choices=["tile_packed_to_4d"],
         required=False,
         help=(
-            "Int4 packing format for encoder linear layers.\n"
+            "Packing format for encoder linear layers.\n"
             "Only applicable to certain backends such as CUDA and Metal\n\n"
             "Options:\n"
             "  tile_packed_to_4d  - int4 4d packing format"


optimum/exporters/executorch/quantization.py
```diff
@@ -83,6 +84,11 @@ def quantize_model_(
                 weight_dtype=torch.int4,
                 granularity=linear_weight_granularity,
             ),
+            "4w_hqq": Int4WeightOnlyConfig(
```
mergennachin: @metascroy @jerryzh168

which config should we use for weight only 4bit (for desktop/laptop on CUDA/Metal backends)?

optimum/commands/export/executorch.py
```diff
@@ -92,14 +93,15 @@ def parse_args_executorch(parser):
...
         required=False,
         help=(
             "Quantization config for linear layers.\n\n"
             "Options:\n"
-            "  8da4w - 8-bit dynamic activation, 4-bit weight\n"
-            "  4w    - 4-bit weight only\n"
-            "  8w    - 8-bit weight only"
+            "  8da4w       - 8-bit dynamic activation, 4-bit weight\n"
+            "  4w          - 4-bit weight only\n"
+            "  4w_hqq      - 4-bit weight and hqq qparams algorithm\n"
```
jackzhxng:
```suggestion
            "  4w_hqq      - 4-bit weight only using hqq qparams algorithm\n"
```

optimum/exporters/executorch/quantization.py
metascroy: How are the CUDA/Metal backends in ET designed?  Are they just a thin wrapper around AOTI, or do they do some graph passes before lowering?

If you use IntxWeightOnlyConfig, you'll get quantization represented as a dequant op followed by a linear op, and this pattern can be matched/consumed by backends.  In this way, a generic quantization config can serve multiple ET backends.  This is how XNNPACK/Vulkan/CoreML work today with quantize_.

And ET's Cuda/Metal backends could theoretically also do this as well.

The code here is not a generic one, but specific to CUDA (see the "tile_packed_to_4d" format).  And after export it will reference CUDA kernels.  It is functional, but is not how other ET backends work.  At the very least, we should put cuda somewhere in the qmode name, e.g., "4w_hqq_cuda".

jerryzh168: the current config looks good, this is targeting tinygemm kernel in CUDA

jerryzh168: for metal, seems like it is calling into the same aten op: https://github.com/pytorch/pytorch/blob/70ec464c1608116df6d379e097f9149b22407456/aten/src/ATen/native/native_functions.yaml#L4218

but we haven't tested this one in torchao

metascroy: I still think if we're taking this route, the qmode needs to reference cuda/metal in some way.  It cannot be a generic "4w_hqq".

We can make qmode="cuda:4w" and qmode="metal:4w" match the same config.

larryliu0820: I think we should take advantage of the optimum-cli `--device` argument. If device is cuda we should set `int4_packing_format="tile_packed_to_4d"`

metascroy: That also works.  In that case, we don't need a new qlinear, but can keep the existing "4w" and have it dispatch to different things based on the device argument.

desertfire: Dispatching is fine, but for CUDA this won't be the only quant format since the current `IntxWeightOnlyConfig` also works for CUDA. I can add another `--qlinear_int4_packing_format` option. How does that sound?

larryliu0820: That works. If we do that, we should error out if user pass in invalid combination like `tile_packed_to_4d` + `xnnpack` etc.

jackzhxng: Having an arg to configure a specify quantization scheme is a bit overkill in my opinion, I'd like to keep the CLI clean so it doesn't become like `export_llama`.

> qmode="cuda:4w" and qmode="metal:4w"

I'd prefer Scott's suggestion here, for other quant formats we express it in the arg e.g. `cuda:4w-<format>`

Either that or we have a more general `qlinear_packing_format` arg and we do some validation that it's used appropriately.

metascroy: Yeah, I'd prefer cuda:4w or metal:4w for qlinear (maybe with -<format>) for these device-specific formats.

> IntxWeightOnlyConfig also works for CUDA

This is true, but using the packed format is more optimized for 4-bit.  You could still access it with "4w" if you want

desertfire: I am not the best person to design what this qmode string should look like, e.g. there is also a `qparams_algorithm` config which potentially can be a part of the qmode string. Maybe there are some other configs that I am not aware of. So I decided to go with adding a `qlinear_packing_format` option. Someone can consolidate things later to add  qmode if they want to.

jackzhxng: @desertfire can you do a general qlinear packing format? `--qlinear_packing_format` instead of `qlinear_int4_packing_format`

optimum/exporters/executorch/quantization.py
```diff
@@ -22,6 +22,7 @@ def quantize_model_(
     eager_model: torch.nn.Module,
     qlinear_config: Optional[str] = None,
     qlinear_group_size: Optional[int] = 32,
+    qlinear_int4_packing_format: Optional[str] = None,
```
larryliu0820: Add `qlinear_encoder_int4_packing_format` maybe? Not critical if you are not using it though.

optimum/exporters/executorch/tasks/causal_lm.py
```diff
@@ -133,8 +133,14 @@ def _load_eager_pretrained(
             param.requires_grad = False
 
     qlinear_config = kwargs.get("qlinear", None)
+    qlinear_int4_packing_format = kwargs.get("qlinear_int4_packing_format", None)
     qembedding_config = kwargs.get("qembedding", None)
-    quantize_model_(eager_model, qlinear_config=qlinear_config, qembedding_config=qembedding_config)
+    quantize_model_(
+        eager_model,
+        qlinear_config=qlinear_config,
+        qlinear_int4_packing_format=qlinear_int4_packing_format,
```
jerryzh168: would it make more sense to merge this into `qlinear_config` itself? and just create a config that's used in torchao quantize_ function? e.g. `Int4WeightOnlyConfig(..., int4_packing_format=qlinear_int4_packing_format)`

jerryzh168: oh nvm, seems like you are doing this in `quantize_model_` itself

optimum/exporters/executorch/quantization.py
desertfire: I am just following the existing convention. In optimum/exporters/executorch/tasks/multimodal_text_to_text.py, quantize_model_ is called separately for the decoder and encoder, so no need to add another `qlinear_encoder_int4_packing_format`.

PR #169, opened 2025-10-16
gemma3 model on et-cuda backend

Gasoonjia:
This PR introduces the exportability of gemma3 model on et-cuda backend.

### DIFF
--- a/install_dev.py
+++ b/install_dev.py
@@ -5,7 +5,7 @@ import sys
 
 def install_torch_nightly_deps():
     """Install torch related dependencies from pinned nightly"""
-    EXECUTORCH_NIGHTLY_VERSION = "dev20251015"
+    EXECUTORCH_NIGHTLY_VERSION = "dev20250916"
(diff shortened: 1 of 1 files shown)


larryliu0820 (approved):

install_dev.py
```diff
@@ -5,7 +5,7 @@
 
 def install_torch_nightly_deps():
     """Install torch related dependencies from pinned nightly"""
-    EXECUTORCH_NIGHTLY_VERSION = "dev20250916"
```
jackzhxng, line 8: Can we pin bump in separate pr?

Gasoonjia: yeah i think we can do that. Although the feature we need for gemma3 model is missing in 0916, the test is not actually running now, so don't pin bump now should be fine.

Himanshunitrr:
Were these merged? I am having problem with gemma3 (https://github.com/huggingface/optimum-executorch/issues/200). Will be very helpful if you could tell what am I doing wrong...

PR #172, opened 2025-10-17
Fix T5

jackzhxng:
Fixes T5 export

Stack from (oldest at bottom):
- https://github.com/huggingface/optimum-executorch/pull/171
- https://github.com/huggingface/optimum-executorch/pull/167
- __->__ https://github.com/huggingface/optimum-executorch/pull/172


optimum/executorch/modeling.py
```diff
@@ -530,9 +527,6 @@ def generate(
         Returns:
             List[int]: List of generated token IDs.
 
-        Note:
-            Temporarily implemented this method in Python due to limited access to ExecuTorch's c++ LLM runner via pybind.
-            Expect improvements to the pybind interface in ExecuTorch version 0.4.1.
```
larryliu0820, line 535: We should use LLM runner

jackzhxng: T5 is not an LLM though

PR #173, opened 2025-10-19
Let ASR models support bfloat16 dtype and cuda device

larryliu0820:
This pull request improves device and dtype handling across the ExecuTorch export pipeline, ensuring that models, tensors, and modules are consistently placed on the correct device and use the appropriate data type. Additionally, it introduces a new test to validate exporting Whisper models with bfloat16 precision and checks the resulting file size.

Device and dtype consistency improvements:

* Updated initialization and export logic in `optimum/exporters/executorch/integrations.py` to use `model.device` and `model.dtype` for all relevant tensors and modules, replacing hardcoded `"cpu"` and `torch.float32` values. This ensures exported models and caches are created on the correct device with the correct data type. [[1]](diffhunk://#diff-a562143f6e3b2d05ff537e787bb475fd109c45be551d2f0d4c3af6696701194dL672-R676) [[2]](diffhunk://#diff-a562143f6e3b2d05ff537e787bb475fd109c45be551d2f0d4c3af6696701194dL738-R738) [[3]](diffhunk://#diff-a562143f6e3b2d05ff537e787bb475fd109c45be551d2f0d4c3af6696701194dL772-R772) [[4]](diffhunk://#diff-a562143f6e3b2d05ff537e787bb475fd109c45be551d2f0d4c3af6696701194dL811-R813) [[5]](diffhunk://#diff-a562143f6e3b2d05ff537e787bb475fd109c45be551d2f0d4c3af6696701194dL825-R827)
* Modified `load_seq2seq_speech_model` in `optimum/exporters/executorch/tasks/asr.py` to accept `device` and `dtype` as keyword arguments, passing them to the underlying model loading and export logic.

Testing enhancements:

* Added a new slow test `test_whisper_large_v3_turbo_export_bfloat16` in `tests/models/test_modeling_whisper.py` to export the Whisper large v3 turbo model with bfloat16 precision, verify the output file exists, and check that its size is approximately 1.2GB with a 10% tolerance.

### DIFF
--- a/optimum/exporters/executorch/tasks/asr.py
+++ b/optimum/exporters/executorch/tasks/asr.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import torchao
-
 from transformers import AutoModelForSpeechSeq2Seq
 
 from ..integrations import Seq2SeqLMExportableModule


optimum/exporters/executorch/integrations.py
```diff
@@ -822,9 +827,15 @@ def export(
             example_encoder_hidden_states = encoder_hidden_states
 
         example_decoder_input_ids = (
-            decoder_input_ids if decoder_input_ids is not None else torch.tensor([[0]], dtype=torch.long)
+            decoder_input_ids
+            if decoder_input_ids is not None
+            else torch.tensor([[0]], dtype=torch.long, device=self.model.device)
```
jackzhxng, line 836: can we add this device piping for other exportable modules so they can be on cuda device as well?

larryliu0820: good idea

PR #177, opened 2025-10-31
Move packaging setup to pyproject.toml + fix installation errors

jackzhxng:
- Moves installation instructions to `pyproject.toml` from legacy `setup.py`
- `pip install .` and later `pip install optimum-executorch` should just work on its own when installed from pypi
- Enables editable install with `pip install -e .`.

### DIFF
--- a/install_dev.py
+++ b/install_dev.py
@@ -5,21 +5,21 @@ import sys
 
 def install_torch_nightly_deps():
     """Install torch related dependencies from pinned nightly"""
-    EXECUTORCH_NIGHTLY_VERSION = "dev20251015"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -1,14 +1,83 @@
...
+  "librosa",
+  "soundfile",
+  "tiktoken",
+  "black~=23.1",
+  "ruff==0.4.4",
+]
+
+[project.urls]
+Homepage = "https://github.com/huggingface/optimum"
+
```
larryliu0820: Point to optimum-executorch?

larryliu0820 (approved):

PR #185, opened 2025-11-11
Update README and fix deps

jackzhxng:


README.md
```diff
@@ -127,7 +122,11 @@ Optimum transformer models utilize:
 - A **custom KV cache** that uses a custom op for efficient in-place cache update on CPU, boosting performance by **2.5x** compared to default static KV cache.
 
 ### Backends Delegation
-Currently, **Optimum-ExecuTorch** supports the [XNNPACK Backend](https://pytorch.org/executorch/main/backends-xnnpack.html) for CPU and [CoreML Backend](https://docs.pytorch.org/executorch/stable/backends-coreml.html) for GPU on Apple devices.
+Currently, **Optimum-ExecuTorch** supports the following backends:
+- [XNNPACK](https://pytorch.org/executorch/main/backends-xnnpack.html) - this is the most supported backend and will work with all models.
+- [CoreML](https://docs.pytorch.org/executorch/stable/backends-coreml.html)
+- Cuda
+- [Metal](https://docs.pytorch.org/executorch/stable/backends/mps/mps-overview.html) - current only available with `executorch >= 1.1.0.dev20251017`. Please separately install this nightly to use the Metal backend.
```
manuelcandales, line 129: This is linking to the MPS backend instead. Also, our instructions are to install executorch from source, not the nightly.

jackzhxng: ET needs to be installed from source for Metal?

manuelcandales: Not particular to Metal. I would say the same for CUDA. Given that we are actively developing these backends, it makes sense to instruct users to install from source (i.e. using the install_executorch.sh script) which is also how things are being tested in CI.

jackzhxng: Makes sense, could you suggest some installation instructions? Do we install ET from source before or after Optimum ET, and do we install from tip of main or a specific commit?

PR #188, opened 2025-11-18
Bump transformers, improve SDPA, reduce Optimum reliance

jackzhxng:
### Summary
Bump transformers to v5.0.0rc1.

### Features
- Bump Optimum to 2.0.0
- Reduce reliance on Optimum functions and classes such as `TaskManager` 1. in preparation for migration to ExecuTorch main repo and 2. since a [fix](https://github.com/huggingface/optimum/pull/2384) for a blocking issue is not yet included in a patch release
- Remove usage of no longer existing `sdpa_without_vmap` for attention and attention mask
- After transformers bump, attention_mask is no longer provided to the attention interface due to causal attention. Rely on `position_ids` instead, which most models seem to already be passing into the interface through kwargs
  - Introduce no-op attention mask implementation since attention mask is only used for sliding window attention, where it is created outside of the attention mask interface
- `get_task` from Transformers is not as robust as `infer_task_from_model` from Optimum, so manually provide task for some models
- Force ExecuTorch dep to use CPU-only torch for testing, this shaves off around 5 GB from test build

### Tests
- Many previously failing tests from no more space on OS are now passing after forcing non-GPU build of ExecuTorch. The rest are running out of memory (`qwen3`, `phi`, `phi4`, `gemma3`) at some point during the testing.
- New failures:
  - `gpt2`: will be fixed upstream by https://github.com/huggingface/transformers/pull/42842
  - `roberta`: https://github.com/huggingface/optimum-executorch/issues/193

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 dependencies = [
   "optimum~=2.0.0",
   "executorch>=1.0.0",
-  "transformers==5.0.0rc0",
+  "transformers==5.0.0rc1",
   "pytorch-tokenizers>=1.0.1",
   "accelerate>=0.26.0",
 ]
--- a/install_dev.py
+++ b/install_dev.py
@@ -35,7 +35,7 @@ def install_dep_from_source():
             "-m",
             "pip",
             "install",
-            "git+https://github.com/huggingface/transformers@cbc6716945cff1d8e124d344ba0150e6e27f8b6e#egg=transformers",  # v5.0.0rc0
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -24,9 +24,9 @@ classifiers = [
 ]
 
 dependencies = [
-  "optimum~=1.24",
+  "optimum~=2.0.0",
   "executorch>=1.0.0",
-  "transformers==4.56.1",
+  "transformers==5.0.0rc0",
```
mergennachin: Looks like rc1 

https://pypi.org/project/transformers/#history

PR #197, opened 2025-12-16
Support torchao MPS 4-bit quantization

manuelcandales:
This pull request extends the quantization and device support for the Executorch export pipeline:
* Added `fpa4w` (floating point activation, 4-bit weight for MPS backend) as a valid choice for the `--qlinear` and `--qlinear_encoder` command-line arguments.
* Added `mps` as a valid choice for the `--device` argument
* Integrated the `UIntxWeightOnlyConfig` from `torchao.experimental.quant_api`.

### DIFF
--- a/optimum/commands/export/executorch.py
+++ b/optimum/commands/export/executorch.py
@@ -221,6 +221,14 @@ class ExecuTorchExportCommand(BaseOptimumCLICommand):
                     "--qlinear_encoder_packing_format can only be used when --qlinear_encoder is set to '4w'"
                 )
 
+        # Validate fpa4w quantization requires MPS device
+        qlinear = getattr(self.args, "qlinear", None)
+        qlinear_encoder = getattr(self.args, "qlinear_encoder", None)
+        if qlinear == "fpa4w" and device != "mps":
+            raise ValueError("--qlinear=fpa4w can only be used when --device is set to 'mps'")
+        if qlinear_encoder == "fpa4w" and device != "mps":
+            raise ValueError("--qlinear_encoder=fpa4w can only be used when --device is set to 'mps'")
+
         kwargs = {}
         if self.args.use_custom_sdpa:
             kwargs["use_custom_sdpa"] = self.args.use_custom_sdpa


optimum/commands/export/executorch.py
```diff
@@ -76,7 +76,7 @@ def parse_args_executorch(parser):
     required_group.add_argument(
         "--qlinear",
         type=str,
-        choices=["8da4w", "4w", "8w", "8da8w", "8da4w,8da8w"],
+        choices=["8da4w", "4w", "8w", "8da8w", "8da4w,8da8w", "fpa4w"],
```
larryliu0820, line 79: Does `fpa4w` work on backends other than metal?

optimum/exporters/executorch/quantization.py
```diff
@@ -101,6 +102,13 @@ def build_linear_config(quant_config_key: str, granularity: str, packing_format:
                     weight_dtype=torch.int8,
                     weight_granularity=PerAxis(0),
                 )
+            if quant_config_key == "fpa4w":
+                # Need to import to load the ops
+                import torchao.experimental.ops.mps  # noqa: F401
```
larryliu0820, line 107: nit should we import this in torchao.experimental.quant_api so that `from torchao.experimental.quant_api import UIntxWeightOnlyConfig` can satisfy the import requirement?

optimum/commands/export/executorch.py
manuelcandales: No, it only works with Metal

larryliu0820: Can you add a check then, if a user pass `--qlinear fpa4w` and `--device mps` at the same time

optimum/exporters/executorch/quantization.py
manuelcandales: `import torchao.experimental.ops.mps` will raise an error if the op library isn't found. The metal ops are not built in torchao by default. For that reason, I thought it would be more clear to have an explicit import that loads the ops, rather than as a side effect of importing the config.

manuelcandales: that's also why I import `torchao.experimental.ops.mps` only if `quant_config_key == "fpa4w"`

optimum/commands/export/executorch.py
manuelcandales: done!

larryliu0820 (approved):

PR #244, opened 2026-07-21
ci: mitigate OOM kills on large-model export tests

JakeStevens:
Large models (e.g. phi4 ~14.7B) exceed the 16GB RAM on standard ubuntu-22.04 runners and get OOM-killed during model load/quantization, surfacing as exit code 143 / 'The operation was canceled' rather than a test failure. Add an aggressive disk cleanup step and a 24GB swapfile so these jobs page to disk instead of being killed, plus a timeout-minutes so genuine hangs fail cleanly instead of dragging the run for hours.

Applied to both the stable (test_models.yml) and nightly (test_models_nightly.yml) workflows.

A larger runner can avoid the need for this swap, but this change keeps it within the current footprint and we can consider a larger runner in the future

### DIFF
--- a/.github/workflows/test_models.yml
+++ b/.github/workflows/test_models.yml
@@ -42,6 +42,9 @@ jobs:
     # Custom job name, now shortened and cleaner
     name: ${{ matrix.test-modeling }} (et=${{ matrix.executorch-version }}, py=${{ matrix.python-version }}, ${{ matrix.os }})
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


digantdesai (approved):

JakeStevens:
I *think* efficientnet failure is due to API request overload, potentially unrelated or maybe uncovered by more of these tests passing and so we have more rpm

Phi4 should be resolved by #245

regisss (approved):

PR #245, opened 2026-07-21
fix(phi4): restore RoPE export workaround for transformers 5.x

JakeStevens:
transformers 5.x renamed the RoPE config field rope_scaling -> rope_parameters (rope_scaling is now a property alias) and dispatches on the 'rope_type' key instead of 'type'. The existing export workaround only set rope_scaling['type'] = 'default', which no longer takes effect: the model still reads rope_parameters['rope_type'] == 'longrope', keeps the dynamic longrope path, and export fails on the data-dependent branch in longrope_frequency_update.

Add a shared helper disable_dynamic_rope_for_export() used by the causal_lm and multimodal_text_to_text tasks. It sets both 'type' and 'rope_type' to 'default' (compatible with transformers 4.x and 5.x), only downgrades the RoPE types with data-dependent branching (longrope, dynamic), and handles the transformers-5 nested per-layer-type rope dict shape.

### DIFF
--- a/optimum/exporters/executorch/tasks/causal_lm.py
+++ b/optimum/exporters/executorch/tasks/causal_lm.py
@@ -20,6 +20,7 @@ from transformers import AutoConfig, AutoModelForCausalLM, GenerationConfig
 from ..integrations import CausalLMExportableModule
 from ..quantization import quantize_model_
 from ..task_registry import register_task
+from ..utils import disable_dynamic_rope_for_export
 
 
 # NOTE: It’s important to map the registered task name to the pipeline name in https://github.com/huggingface/transformers/blob/main/utils/update_metadata.py.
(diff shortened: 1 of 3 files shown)


regisss:
We may still be getting an OOM error with Phi: https://github.com/huggingface/optimum-executorch/actions/runs/29868750873/job/88852447419?pr=245

JakeStevens:
Yes I've put up:

https://github.com/huggingface/optimum-executorch/pull/246

to try to address that. Seems like /mnt has variable amount of space, I added fallbacks to try /, and to more dynamically size the swap

regisss (approved):
LGTM!
I let you merge the PR since you have write rights now :)

JakeStevens:
all green, beautiful :)

PR #246, opened 2026-07-22
Allocate only spare capacity for swap and fail only on OOM

JakeStevens:
We now  size the swapfile to whatever's free (or skips with a ::warning::), so the swap step no longer aborts the job


JakeStevens:
sorry for the churn. This *seems* more stable, but still seems like runner just does not have enough memory at times; need to consider still if we can have larger runners for larger models.

regisss:
> need to consider still if we can have larger runners for larger models.

Let me see what we can do

regisss (approved):

